PR #45, opened 2024-04-14
Added Switch Button to the GtkHelper

G4PLS:
Added a simple Gtk Switch for the Plugin Actions.
The user can specify the title that gets displayed on the Left Hand side.
If the button is active or not needs to be done separate because passing it in as a required/optional argument seemed counter productive.

How a Switch Button would look in an Plugin
image

### DIFF
--- a/GtkHelper/GtkHelper.py
+++ b/GtkHelper/GtkHelper.py
@@ -437,8 +437,19 @@ class ScaleRow(Adw.PreferencesRow):
         self.main_box.append(self.scale)
         self.main_box.append(self.label_right)
 
+class SwitchRow(Adw.PreferencesRow):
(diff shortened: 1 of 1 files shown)


Core447:
Nice, but [Adw.SwitchRow](https://gnome.pages.gitlab.gnome.org/libadwaita/doc/1-latest/class.SwitchRow.html) is basically the exact same thing. You can check what is already available on [Gtk](https://docs.gtk.org/gtk4/visual_index.html) and [Adw](https://gnome.pages.gitlab.gnome.org/libadwaita/doc/main/widget-gallery.html)

G4PLS:
Oh, didnt see that this is already existing
Alright that makes it a lot easier

PR #51, opened 2024-04-16
Created first implementation for Plugin Events

G4PLS:
Plugins can now Create Custom Events just like ActionHolders.
Events can be triggered by the Plugin itself by providing the specified ID
Events follow the Conventions of ActionHolders (An Event ID has to be specified)

The Events themselves implement the Observer pattern in an asynchronous way.
After a callback got connected to the Event, when it triggers the callback will be called async, when the callback is not specified as async a thread will be used to run the callback.
When this fails an error gets thrown and a async wait with a delay of 0 gets used instead to not disrupt the flow of the program.

I found no problems by always creating a new event loop, bigger scale tests may need to be run. Reverting the Changes is easy when it disrupts the program to much

### DIFF
--- a/src/backend/PluginManager/EventHolder.py
+++ b/src/backend/PluginManager/EventHolder.py
@@ -1,0 +1,43 @@
+import asyncio
+from loguru import logger as log
+
+class EventHolder:
+    """
+        Holder for Event Callbacks for the specified Event ID
+    """
+    def __init__(self, plugin_base: "PluginBase", event_id: str):
+        if event_id in ["", None]:
+            raise ValueError("Please specify an signal id")
+
+        self.plugin_base = plugin_base
+        self.event_id = event_id
+        self.observers: list = []
+
+    def add_listener(self, callback: callable):
+        if callback not in self.observers:
+            self.observers.append(callback)
+        else:
+            log.warning(f"Callback {callback.__name__} is already subscribed to: {self.event_id}")
+
+    def remove_listener(self, callback: callable):
+        if callback in self.observers:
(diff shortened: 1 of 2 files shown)


Core447:
Maybe add a method to `PluginBase` which allows to call the events of other plugins

G4PLS:
> Maybe add a method to `PluginBase` which allows to call the events of other plugins

Currently I have it so that you can only subscribe to events and not trigger events of other Plugins
My intention was that the Plugin creator still has a little bit more flexibility.
Maybe someone creates an event that is mostly used internally for the Plugin and you dont want an external Plugin to trigger it. Currently it would be harder externally, what i was going for.

I could create a separation between Public and Private events tho

Also how can you currently get the active actions so you can subscribe to events? I think you talked about it at one point

Core447:
> Also how can you currently get the active actions so you can subscribe to events? I think you talked about it at one point

That's not directly possible, but I think the best way would be to have a list of all actions in the PluginBase where the actions add themselves to.

G4PLS:
> > Also how can you currently get the active actions so you can subscribe to events? I think you talked about it at one point
> 
> That's not directly possible, but I think the best way would be to have a list of all actions in the PluginBase where the actions add themselves to.

Can you get the Loaded Plugins?
So you basically get the PluginBase of a Plugin you want by providing functions to do so
Because when you can get the PluginBase you have access to basically anything you need of the Plugin

When there currently is nothing like that I will take a look how to do that. Currently Events would then only work on a per-plugin basis

PR #118, opened 2024-05-25
Update French locale

mrintrepide:
Add and fix French locale


Core447:
Great work! Thank you!!

PR #122, opened 2024-05-30
fix: show_error is on ControllerKey, not ControllerKeyState

Sorunome:


Core447:
Thanks for the great work! However, it might make more sense to actually handle the error icon separately for all states, what do you think?

Sorunome:
While that does make sense, it also sounds like a different thing that this pr does; this pr fixes the status quo, the other thing would be an enhancement

30. 5. 2024 20:54:38 Core447 ***@***.***>:

Sorunome:
Superseded by #124, there it is also on the state

Core447:
Ok, then I'm going to add this to the state

PR #124, opened 2024-05-31
Soru/steam deck plus

Sorunome:


Sorunome:
Known issues:
 - Have to restart after deleting an action, no idea why

Sorunome:
Soru only added dial stuffs nicely, for touchscreen stuffs (setting image, actions, etc) soru'd like help, if possible. Or mayhaps implement it in a future PR

Core447:
I finally went through the code. A "couple" of things I noticed:
- maybe combine type and identifier by either passing the key/dial/touchscreen object or by using an enum - the same goes for action_support of ActionHolder (maybe convert it to a dict?)
- ActionBase: combine get_xxx_state in one method get_state and remove get_input_state because one action can only be on one element at a time
- I'm going to update the callback methods to enums
- we need a method to set the touchscreen image for actions on the dials
- allow show and hide_error on multiactions if it has the image permission (not directly related to your pr)
- ActionBase: the `if not text: return` blocks plugins from clearing the labels
- TODO in StoreBackend: maybe move this to the plugin manager - something like remove_all_actions_of_plugin or to the Page and let the plugin manager just iterate over the pages and call their methods
- The name type that you use may conflict with the python method type later on - we need to find another name for it
- maybe change sidebar active_type to the singular form?
- maybe other system for the action_objects than the dict (not directly related to your pr)
- has_key_an_image_controll_action is used in ActionConfigurator even though you pass in the type - bit weird to read maybe we can abstract it
- load_for_screen in ActionManager now used the identifier instead of gesture, even though it shouldnt. Identifier is the nth screen where gesture is swipe-left, swipe-right (will be replaced by an enum). However, it might make sense to remove the gestures as you did and let users use a plugin like AdvancedActions from @G4PLS  if they want to have individual actions
- We can now remove the ActionSupports class
- The support for all types should be UNTESTED by default

Again, amazing work!

Sorunome:
> maybe combine type and identifier by either passing the key/dial/touchscreen object or by using an enum - the same goes for action_support of ActionHolder (maybe convert it to a dict?)

Do you mean something like `Input.Key("0x2")` and `Input.Dial("4")` and `Input.Touchscreen("touchscreen")` and pass that around?

> we need a method to set the touchscreen image for actions on the dials

Yeah, that is something soru did not include in this PR as soru noticed that things with updating images were getting a bit over her head

> maybe change sidebar active_type to the singular form?

What do you mean?

> load_for_screen in ActionManager now used the identifier instead of gesture, even though it shouldnt. Identifier is the nth screen where gesture is swipe-left, swipe-right (will be replaced by an enum). However, it might make sense to remove the gestures as you did and let users use a plugin like AdvancedActions from @G4PLS if they want to have individual actions

Yeah, soru was thinking of something like that

> We can now remove the ActionSupports class

No? we still have to filter out which actions are available on which input method?


Thank you for the feedback, will get to coding :)

Core447:
> Do you mean something like `Input.Key("0x2")` and `Input.Dial("4")` and `Input.Touchscreen("touchscreen")` and pass that around?

Yes

> Yeah, that is something soru did not include in this PR as soru noticed that things with updating images were getting a bit over her head

No problem, I can do this and also wallpapers etc for the screen(s).

> What do you mean?

`active_type` of Sidebar can be `dials`, `touchscreen`, keys`. I think it makes more sense to rename that to `dial`, `touchscreen`, `key` because you can only activate one key, dial or screen at a time.

> Yeah, soru was thinking of something like that

Of what exactly? To rely on a plugin if a user wants to specify it individually?

> No? we still have to filter out which actions are available on which input method?

I thought you replaced it by `TypeSupport`, which also makes more sense than `ActionSupport`?

PR #130, opened 2024-06-12
Image and label permissions are now stored in each action instead of relying on an index

gensyn:


Core447:
I'm waiting with the merge until the stream-deck-plus branch is finished and merged

Core447:
Tbh I prefer the way current way of storing the permissions, but I see your point: Your approach is less likely to cause problems. Can you please check if you're still experiencing this issue on the head of main (changed some stuff)?

PR #134, opened 2024-06-17
Use fallback locale when can't parse locale

ulrikstrid:
When OS locale is invalid everything breaks so we should use the fallback when it comes back as None


Core447:
Thank you!

PR #135, opened 2024-06-17
Replace self.HOLD_TIME with self.deck_controller.hold_time

ulrikstrid:
self.HOLD_TIME is not defined and it seems like we should use the 'global' value stored on the deck controller anyway


Core447:
Thanks! Seems like I forgot to change that after my refactor.

PR #136, opened 2024-06-17
Handle when row.event is None

ulrikstrid:
row.event is sometimes None, when that happens we get a error because we can't use  on None. Fixed by checking if it's None before using


Core447:
Thanks for reporting! However, this just avoids the error, the real problem is that `row.event` can be `None`. d1400e1e1910eb9660e6c8483e30dc3702287a91 Should fix this.

PR #145, opened 2024-06-24
Feat(ui): add custom ComboRow wrapper for a simple ComboBox with static options

GsakuL:
while working on StreamController/Requests#2 it became clear, that a "combo box" style widget is not as easy as I thought.
Here is a pretty basic custom widget to wrap a few Adw/Gtk classes to provide a combo box for static items.
Items have keys to be stored in persistent settings, and have names, that can be localized.

- I wasn't sure who the 'Author' should be in the header, if it's you as the maintainer ("author of the project"), or me as the "author of the file".
- I could not find an already existing solution in the project, please excuse if there is
- I thought `widgets/` would be appropriate since it's intended to be "re-used anywhere"
- proper Enums and generics `[T]` would be better, but `GEnum` currently seems to be not supported, thus i though str keys are the most flexible and "user friendly" (instead of cryptic int values) for now
- as I did not write the core functionality, I put a link for a lack of better attribution 
- it would be nice to have this in the next release (so the Requests can depend on that), but I don't want to be pushy

(dirty) Example:
```py
def get_config_rows(self):
    _receive_types = [
        ListItem(k, lm.get(f"convert.list_item.{k}"))
        for k in ["json", "xml", "plain_text", "ignore"]
    ]
    self.receive_type = StaticItemListComboRow(
        _receive_types,
        title=lm.get("actions.get.receive_type.title"))
    self.receive_type.connect("notify::selected", self.on_receive_type_changed)

def on_receive_type_changed(self, entry, *args):
    settings = self.get_settings()
    settings["receive_type"] = entry.get_selected_item().key
    self.set_settings(settings)
```

and it would look like this (with localization)
grafik

Edit: could not get screenshots with mouse pointer and mouse over-effect, so here a small clip:
[example.webm](https://github.com/StreamController/StreamController/assets/6844160/c3da194d-6e0a-484e-8946-f94760853dab)

### DIFF
--- a/GtkHelper/StaticItemListComboRow.py
+++ b/GtkHelper/StaticItemListComboRow.py
@@ -1,0 +1,87 @@
+"""
+Author: GsakuL
+Year: 2024
+
+This program is free software: you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation, either version 3 of the License, or
+any later version.
+
+This programm comes with ABSOLUTELY NO WARRANTY!
+
+You should have received a copy of the GNU General Public License
+along with this program. If not, see <https://www.gnu.org/licenses/>.
+"""
+
+from typing import Sequence
+
+import gi
+gi.require_version("Gtk", "4.0")
+gi.require_version("Adw", "1")
+from gi.repository import Gtk, Adw, Gio, GObject
+
+from loguru import logger as log
+
+
+class ListItem(GObject.Object):
+    def __init__(self, key: str, name: str):
+        super().__init__()
+        self._key = key
+        self._name = name
+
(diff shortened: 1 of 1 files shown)


Core447:
Great work, thank you! This will definitely be included with the next release.
Feel free to write your (user-)name into the file because you're the author of it.
And maybe we can move this file into `GtkHelper` (not directly `GtkHelper.GtkHelper.py`, just the folder `GtkHelper`, what do you think? This is where all the other generic helper widgets are.

GsakuL:
`GtkHelper` folder sounds just as good. I'll make the changes

Core447:
Perfect, thank you!

PR #147, opened 2024-06-28
feat: add ability to use line breaks in labels

axolotlmaid:
## Changes
- Changed Entry to TextView in LabelEditor.py
- Removed placeholder text
- Added CSS entries to make the TextView fit in

## Images
2024-06-28-221936_hyprshot
2024-06-28-221923_hyprshot

### DIFF
--- a/locales/locales.csv
+++ b/locales/locales.csv
@@ -8,6 +8,7 @@ label-editor-bottom-name;Unten;Bottom;Bas
 label-editor-font-chooser-label;Schriftart:;Font:;Police :
 label-editor-font-weight-label;Schriftdicke:;Font weight:;Taille de la police:
 label-editor-expander-subtitle;Texte für diese Taste;Labels for this key;Étiquettes de ce bouton
+label-editor-placeholder-text;Text;Label;Étiquette;Etiqueta
 action-editor-header;Aktions;Actions;Actions
 action-editor-expander-subtitle;Aktions für diese Taste;Actions for this key;Actions pour ce bouton
 action-editor-add-new-action;Aktion hinzufügen;Add Action;Ajouter une action
--- a/locales/locales.csv
+++ b/locales/locales.csv
@@ -7,7 +7,6 @@ label-editor-center-name;Mitte;Center;Centre
 label-editor-bottom-name;Unten;Bottom;Bas
 label-editor-font-chooser-label;Schriftart:;Font:;Police :
 label-editor-font-weight-label;Schriftdicke:;Font weight:;Taille de la police:
-label-editor-placeholder-text;Text;Label;Étiquette
 label-editor-expander-subtitle;Texte für diese Taste;Labels for this key;Étiquettes de ce bouton
 action-editor-header;Aktions;Actions;Actions
 action-editor-expander-subtitle;Aktions für diese Taste;Actions for this key;Actions pour ce bouton
--- a/locales/locales.csv
+++ b/locales/locales.csv
@@ -9,6 +9,7 @@ label-editor-font-chooser-label;Schriftart:;Font:;Police :
 label-editor-font-weight-label;Schriftdicke:;Font weight:;Taille de la police:
 label-editor-expander-subtitle;Texte für diese Taste;Labels for this key;Étiquettes de ce bouton
 label-editor-placeholder-text;Text;Label;Étiquette;Etiqueta
+label-editor-color-chooser-label;Farbe:;Color:;Couleur:
 action-editor-header;Aktions;Actions;Actions
 action-editor-expander-subtitle;Aktions für diese Taste;Actions for this key;Actions pour ce bouton
 action-editor-add-new-action;Aktion hinzufügen;Add Action;Ajouter une action
--- a/src/windows/mainWindow/elements/Sidebar/elements/LabelEditor.py
+++ b/src/windows/mainWindow/elements/Sidebar/elements/LabelEditor.py
@@ -125,17 +125,19 @@ class LabelRow(Adw.PreferencesRow):
                                                     margin_bottom=3, visible=False)
         self.main_box.append(self.controlled_by_action_label)
 
-        self.text_box = Gtk.Box(orientation=Gtk.Orientation.HORIZONTAL, hexpand=True)
-        self.main_box.append(self.text_box)
-
         self.text_view = TextView(placeholder=gl.lm.get("label-editor-placeholder-text"))
-        self.text_view.set_margin_end(5)
         # self.text_view.entry.connect("changed", self.on_change_text)
-        self.text_box.append(self.text_view)
+        self.main_box.append(self.text_view)
+
+        self.color_chooser_box = Gtk.Box(orientation=Gtk.Orientation.HORIZONTAL, hexpand=True, margin_top=6)
+        self.main_box.append(self.color_chooser_box)
+
+        self.color_chooser_label = Gtk.Label(label=gl.lm.get("label-editor-color-chooser-label"), xalign=0, hexpand=True, margin_start=2)
+        self.color_chooser_box.append(self.color_chooser_label)
 
         self.color_chooser_button = ColorChooserButton()
         # self.color_chooser_button.button.connect("color-set", self.on_change_color)
-        self.text_box.append(self.color_chooser_button)
+        self.color_chooser_box.append(self.color_chooser_button)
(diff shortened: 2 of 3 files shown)


Core447:
Nice work! I'll try this when I have some time

Core447:
I changed the design a bit. The only thing bothering me is that the TextView has no placeholder support...

axolotlmaid:
> I changed the design a bit. The only thing bothering me is that the TextView has no placeholder support...

Yeah, that is why i had to remove the placeholder text. You can technically add line breaks to a label without a TextView but I think it looks a bit weird where instead of adding new lines in the Entry widget it just adds this weird arrow symbol.

Core447:
I'll play around with different concepts for this when I have some time

gensyn:
> I'll play around with different concepts for this when I have some time

Do you have a branch with your changed design? I don't see a fittting name in the StreamController branches.

Core447:
So far it's just the one commit (fb00101a0ffd4300268d5e11e80e00cc96108dfc) that I made to this PR. I definately want to keep some sort of placeholder. What do you think about simply allowing \n in the currently used entry box?

gensyn:
> So far it's just the one commit ([fb00101](https://github.com/StreamController/StreamController/commit/fb00101a0ffd4300268d5e11e80e00cc96108dfc)) that I made to this PR. I definately want to keep some sort of placeholder. What do you think about simply allowing \n in the currently used entry box?

Would be fine by me. Though I would like to mention that I asked ChatGPT about the placeholder issue with TextView and it suggested to handle it ourselves like this:

```python
import gi
gi.require_version('Gtk', '3.0')
from gi.repository import Gtk, Gdk

class MainWindow(Gtk.Window):
    def init(self):
        Gtk.Window.init(self, title="TextView with Placeholder Example")
        self.set_default_size(400, 300)

        # VBox to arrange widgets vertically
        vbox = Gtk.VBox(spacing=6)
        self.add(vbox)

        # Button to replace the entry with a TextView
        self.button = Gtk.Button(label="Replace with TextView")
        self.button.connect("clicked", self.on_button_clicked)
        vbox.pack_start(self.button, False, False, 0)

        # Initial single-line text entry
        self.entry = Gtk.Entry()
        self.entry.set_text("This is a single-line entry")
        vbox.pack_start(self.entry, False, False, 0)

        # Container for the TextView (will be shown later)
        self.textview_scrolled_window = Gtk.ScrolledWindow()
        self.textview = Gtk.TextView()
        self.textview_scrolled_window.add(self.textview)
        self.textview_scrolled_window.set_vexpand(True)

        # Placeholder text
        self.placeholder_text = "Enter your text here..."
        self.is_placeholder_visible = False

    def on_button_clicked(self, widget):
        # Transfer text from Entry to TextView
        entry_text = self.entry.get_text()
        text_buffer = self.textview.get_buffer()
        text_buffer.set_text(entry_text)

        # Remove Entry and show TextView in the scrolled window
        parent = self.entry.get_parent()
        parent.remove(self.entry)
        parent.pack_start(self.textview_scrolled_window, True, True, 0)
        self.textview_scrolled_window.show_all()

        # Connect signals for placeholder management
        self.textview.connect("focus-in-event", self.on_textview_focus_in)
        self.textview.connect("focus-out-event", self.on_textview_focus_out)

        # Set placeholder if TextView is empty
        if text_buffer.get_text(text_buffer.get_start_iter(), text_buffer.get_end_iter(), True) == "":
            self.set_placeholder()

        # Disable the button since replacement is already done
        self.button.set_sensitive(False)

    def set_placeholder(self):
        buffer = self.textview.get_buffer()
        buffer.set_text(self.placeholder_text)
        self.textview.modify_fg(Gtk.StateFlags.NORMAL, Gdk.color_parse("gray"))
        self.is_placeholder_visible = True

    def remove_placeholder(self):
        buffer = self.textview.get_buffer()
        buffer.set_text("")
        self.textview.modify_fg(Gtk.StateFlags.NORMAL, None) # Reset to default color
        self.is_placeholder_visible = False

    def on_textview_focus_in(self, widget, event):
        if self.is_placeholder_visible:
            self.remove_placeholder()

    def on_textview_focus_out(self, widget, event):
        buffer = self.textview.get_buffer()
        text = buffer.get_text(buffer.get_start_iter(), buffer.get_end_iter(), True)
        if text.strip() == "":
            self.set_placeholder()

win = MainWindow()
win.connect("destroy", Gtk.main_quit)
win.show_all()
Gtk.main()
```

Core447:
That doesn't work for me, I just get a white window. (and it's gtk3)

gensyn:
Well yeah, ChatGPT rarely spits out code that just works. It was more meant as concept.

Put in grey placeholder text by default and when entering the TextView check if the text is the placeholder - if yes, remove it. On leaving the TextView, check whether it is empty and if yes, readd the placeholder.

Core447:
> Put in grey placeholder text by default and when entering the TextView check if the text is the placeholder - if yes, remove it. On leaving the TextView, check whether it is empty and if yes, readd the placeholder.

But if the user want to have the placeholder text as the actual text...

gensyn:
:laughing: 
I feel like having
```
Label line 1
Label line 2
```
as a placeholder is a safe bet. As a workaround, people who reeeealy want that text could add an extra space.

Core447:
I hope I'm not annoying, but I really don't want to just bet that no one will use the placeholder. If there's no way to add a placeholder via some workaround, I think we should use the `\n` option

gensyn:
By now it seems like you are looking for reason not to use TextView.
If you wanted to, you could save a boolean if the text in the TextView was set automatically or by the user and there are probably a couple of other solutions as well. And as I said, I don't really care either way

Core447:
I don't have anything against TextView, but I hope you can understand that I want to keep the placeholder. However, I'll try to find a nice workaround for the placeholder

Core447:
@gensyn What do you think about this placeholder solution?

gensyn:
I played around with it a bit and it seems to be working good. Once I got to a point to where the TextView would only display one line although there were several but I couldn't reproduce it. Looks good.

gensyn:
But maybe the color button should have a fixed height - that looks a little weird when it stretches.

Core447:
> But maybe the color button should have a fixed height - that looks a little weird when it stretches.

image
Without stretching it looks weird too...

gensyn:
True.

Core447:
> True.

And while the stretched button also looks kinda weird, it at least makes sense, because it controls the color of all lines. However, maybe moving it below the textview makes sense

gensyn:
Then it could better line up with the new outline row. Sounds like a good idea.

gensyn:
I was actually thinking that font and color could share a line like oultine width and outline color do. That way, both color pickers would be next to each other. I thought that might look tidier.

Core447:
I think I would make every property take one line, separating outline width and outline color. What do you think?

gensyn:
Yeah, sure, as long as it's consistent.

gensyn:
That works for me.

Core447:
It takes up quite a lot of space, but it's consistent now.

There's just one problem left: When there's a multi-line label in the textview and the revert/reset button is pressed the text gets cleared as expected, but the height doesn't shrink.

gensyn:
Not always, but sometimes the height does get reset. I assume it's a race condition between deleting the text and the focus lost event of the TextView or something like that.

gensyn:
I think this a minor cosmetic problem which wouldn't bother me if everything else works. So I would just go with it instead of leaving this PR in limbo.

Core447:
The problem is that users will think it's a bug, which it kinda is, so I'll try to fix this in coming days and if I can't fix it, I'm gonna merge it

Core447:
I still haven't found a way around this, what do you think about a toggle for this feature in the settings so that the users will get informed about this "bug" and don't report it?

gensyn:
Yeah, that sounds reasonable. Although I suspect that instead user will keep asking for this feature because it's hidden. 🙃

axolotlmaid:
> Yeah, that sounds reasonable. Although I suspect that instead user will keep asking for this feature because it's hidden. 🙃

What about adding an experimental section in the first start up page?

Core447:
> What about adding an experimental section in the first start up page?

The problem with this is that it can't be changed later on because there's no way to open the onboarding dialog a second time.

PR #173, opened 2024-07-18
Added default font for labels

gensyn:
When running StreamController I get a couple of log entries like:
`findfont: Font family [''] not found. Falling back to DejaVu Sans.`

I'm not exactly sure which labels are causing this, but I thought since DejaVu Sans is the fallback anyway, we might as well make it the default. Might be different on other systems but it should not break anything.

### DIFF
--- a/src/backend/DeckManagement/DeckController.py
+++ b/src/backend/DeckManagement/DeckController.py
@@ -1210,7 +1210,7 @@ class LabelManager:
         if label.color is None:
             label.color = [255, 255, 255, 255]
         if label.font_name is None:
-            label.font_name = ""
+            label.font_name = "DejaVu Sans"
         if label.font_size is None:
             label.font_size = 15
 


Core447:
Ok, I just want to check what happens if DejaVu Sans is not installed on the system. Maybe we can also have a list of fallback options and we use the first one available on the system.

gensyn:
Well, as it is now, the fallback has been handled by PIL (I think) so not much can change. If DejaVu Sans is not installed, the behavior should be the same as before.

gensyn:
It wasn't PIL, it was line 34 in KeyLabel.py: `matplotlib.font_manager.FontProperties(family=self.font_name)`.

Core447:
> It wasn't PIL, it was line 34 in KeyLabel.py: `matplotlib.font_manager.FontProperties(family=self.font_name)`.

Yeah, and I assume that matplotlib always chooses an available font, but I'm not sure

Core447:
I've a method that tries to use DejaVu but can fallback to other system fonts it it doesn't exist in 8c8d754edd1c4f9b7a6a3108d936f1006a56f3e8.

PR #174, opened 2024-07-18
Added option to change the outline color of labels

gensyn:


Core447:
Great job! We just need a revert button for the outline width spinner (similar to the size spinners in the layout area). And we should add arguments to `set_label` of `ActionBase` to set defaults for the new options

gensyn:
On it.

gensyn:
Ok, added both.

Core447:
Thank you!

gensyn:
Thanks for adding the last tweaks

Core447:
No problem!

PR #179, opened 2024-07-18
fix: icon store page(s) showing error

axolotlmaid:
The `description` variable is returning a `dict` type. An example of the dict is below:
> {'en_US': 'Unofficial - Material Icons designed by Google.'}

This pull request changes the code to set the label to the user's locale from the dictionary, if it is found.

I also added a fallback because in the plugins tab of the store it returns a `string` instead of a `dict`.

### DIFF
--- a/src/windows/Store/Icons/IconPage.py
+++ b/src/windows/Store/Icons/IconPage.py
@@ -123,7 +123,7 @@ class IconPreview(StorePreview):
 
         # Update info page
         self.icon_page.info_page.set_name(self.icon_data.icon_name)
-        self.icon_page.info_page.set_description(self.icon_data.descriptions)
+        self.icon_page.info_page.set_description(gl.lm.get_custom_translation(self.icon_data.descriptions))
         self.icon_page.info_page.set_author(self.icon_data.author)
         self.icon_page.info_page.set_version(self.icon_data.icon_version)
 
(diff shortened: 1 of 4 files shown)


Core447:
Thank you! However, the description should already get translated by the following lines:
https://github.com/StreamController/StreamController/blob/8c8d754edd1c4f9b7a6a3108d936f1006a56f3e8/src/backend/Store/StoreBackend.py#L328-L329
and
https://github.com/StreamController/StreamController/blob/8c8d754edd1c4f9b7a6a3108d936f1006a56f3e8/src/windows/Store/Plugins/PluginPage.py#L128

Maybe I'm overlooking something, if so, please let me know

axolotlmaid:
> Thank you! However, the description should already get translated by the following lines:
> 
> https://github.com/StreamController/StreamController/blob/8c8d754edd1c4f9b7a6a3108d936f1006a56f3e8/src/backend/Store/StoreBackend.py#L328-L329
> 
> and
> 
> https://github.com/StreamController/StreamController/blob/8c8d754edd1c4f9b7a6a3108d936f1006a56f3e8/src/windows/Store/Plugins/PluginPage.py#L128
> 
> Maybe I'm overlooking something, if so, please let me know

https://github.com/StreamController/StreamController/blob/3c13e6efbd1b1241544a9bc438948a3d5e725bb6/src/windows/Store/Icons/IconPage.py#L126

It seems like you forgot to update this for `IconPage.py`. This actually makes the fix a lot more easier by changing only one line which you can find in the new commit.

Core447:
We can even use the pre-translated variable `description` of the data (see d2be101f28ab1e1365aad1a7defbd655ef43c4ab and 5881840d10b666ababdfa9fdd624bec3eae5bfb6).

Core447:
Thank you!

PR #184, opened 2024-07-24
Added better Pack validation

G4PLS:
Added is_valid fields to packs that gets set to False if a path is not valid
Only adds Packs to the dict when the pack is valid, "removing" unvalid packs from the preview
Reverting the Posix paths pack to a string when creating the preview


Core447:
I didn't go through all the details yet, but it looks very nice!

Thank you!

Core447:
I hope my delay isn't the reason for the closure of this PR... I wasn't home for the last days

G4PLS:
Ohhh, I think it did that because of the other PR
I have to redo this and will create another PR

PR #186, opened 2024-07-28
Feat(GtkHelper): Added an IpEntryRow

G4PLS:
This makes it easier for the User to enter an Ip Address. It does not check if an ip address is valid, this needs to be done seperately

The row itself looks like this:
image

It implements a custom event called 'ip-changed' which triggers whenever any of the entries get changed. This event has 2 args:
- arg1: Event Caller
- arg2: New IP Address as a string

It also has a custom type called 'IpEntryRow'


G4PLS:
Split the Row up and added a NetworkRow containing the IpEntryRow and HostnameEntryRow
Create a IpEntryRow and a HostnameEntryRow which contain the Ui and methods to work out of the box.

Made the HostnameEntryRow only emit the changed signal when enter is pressed and not on every letter that gets changed. Wanted to add focus-out-event aswell so it gets emitted when the user is "finished" but event somehow doesnt exist.

The whole row now looks like this:
image

Core447:
Great work! I'm wondering, though: What is the benefit over an EntryRow? I don't want to sound mean, I just want to know how you came up with the idea.

G4PLS:
I personally think it looks a lot cleaner so its more of a Visual decision i made
I want to add some features like auto focus swapping when entering a Dot

G4PLS:
I created a new branch and when the features are finished would love to merge from there. Wanted to keep my main free and wasnt able to change it in here.
New PR is #190

PR #187, opened 2024-07-30
Fixed StreamDeck not reconnecting

gensyn:
At least on my system (Ubuntu 24.04), my Stream Deck Mini identifies with vendor name "Elgato_Systems", not "Elgato". So the `on_connect()` and `on_disconnect()` functions were never executed for me. I added "Elgato_Systems" as a valid vendor and now unplugging and reconnecting the StreamDeck works as expected.

Alternatively, you could just check for the string "Elgato" in the vendor name or check the actual vendor id, which should be identical on all systems.

### DIFF
--- a/src/backend/DeckManagement/DeckManager.py
+++ b/src/backend/DeckManagement/DeckManager.py
@@ -42,7 +42,7 @@ from gi.repository import GLib, Xdp
 # Import globals
 import globals as gl
 
-VALID_VENDORS = ["Elgato", "Elgato_Systems"]
+ELGATO_VENDOR_ID = "0fd9"
 
 
 class DeckManager:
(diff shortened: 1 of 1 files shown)


Core447:
> Alternatively, you could just check for the string "Elgato" in the vendor name or check the actual vendor id, which should be identical on all systems.

I think this approach makes more sense because we don't know which other strings are used. However, using the "actual" vendor id should be sufficient, isn't it?

gensyn:
I would think so, yes. The vendor ID should be supplied by the StreamDeck itself and therefore always be the same.

Core447:
Ok, do you want to implement this, or should I do it?

gensyn:
I can gladly change my PR on monday, if you dont't want it asap. :)

Core447:
No need to hurry

gensyn:
Ok, I changed the PR to use the vendor ID `0fd9`.

Core447:
Great, thank you!

PR #195, opened 2024-08-06
Housekeeping: keep page backups for one week at most

gensyn:
As the backup count for pages quickly rises (I have deleted several hundred backup folders; power users probably have more), I have added a mechanism to clean them up. Backup folders are now deleted after one week - but at least three backup folders are always kept.

### DIFF
--- a/main.py
+++ b/main.py
@@ -122,6 +122,7 @@ def create_global_objects():
     gl.media_manager = MediaManager()
     gl.asset_manager_backend = AssetManagerBackend()
     gl.page_manager = PageManagerBackend(gl.settings_manager)
+    gl.page_manager.remove_old_backups()
     gl.page_manager.backup_pages()
     gl.icon_pack_manager = IconPackManager()
     gl.wallpaper_pack_manager = WallpaperPackManager()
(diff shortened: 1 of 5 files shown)


Core447:
Thanks!

What do you think about this? I think having the removal code in the PageManagerBackend makes a lot more sense, because this is where they get generated and checking for old backups for each page backup is very resource heavy.

gensyn:
To be honest, I was a little confused by which backups were created where. As long as it works now I'm fine. Code looks good.

gensyn:
But you are just keeping a fixed number of backups, right? No check for the age of a backup?

Core447:
> But you are just keeping a fixed number of backups, right? No check for the age of a backup?

Yes, what is the advantage of checking the age? In theory only keeping one backup should be enough, but just to be sure I set it to 5 (+1).

gensyn:
Well, you suggested one week as a time period when we talked about it on Discord so I just went with it since it's definitely better than the status quo. But it's even better now so I'm all for merging.

Core447:
Ok, sorry, I forgot I said that.

PR #199, opened 2024-08-12
Add support for screensaver under Cinnamon

### LINKED ISSUE #198: Add support for Cinnamon desktops for locking Stream deck
StreamController supports locking the Stream deck when the system's screen is locked.  The Cinnamon desktop on Linux Mint is mostly GNOME, but doesn't work as it.  Request that this be added.

The dbus ScreenSaver stuff in Cinnamon is almost the same as GNOME, but some of the paths/names are changed.

(pull request coming soon)

jfbauer432:
This change adds support for locking the Stream Deck when run under the Cinnamon desktop environment.  Tested on Linux Mint 22 with Xorg.

This probably needs an additional change to work with Wayland/Cinnamon whenever that gets usable (I could not even lock the screen on Linux Mint Cinnamon Wayland when I tried it so they got some work to do yet).

closes #198

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ Available for GNOME, Hyprland and all X11 desktops, this feature allows you to a
 
 ## Auto-Lock
 
-Lock your Stream deck when your system is locked, preventing unwanted use from third parties (available on KDE and GNOME).
+Lock your Stream deck when your system is locked, preventing unwanted use from third parties (available on KDE and GNOME, and Cinnamon).
 
 ## Installation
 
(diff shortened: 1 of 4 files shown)


Core447:
Thank you, great work! It should work under Wayland as well, but I/we need to check this before meeting it. I won't be home for the next ~ week, sorry for the delay.

jfbauer432:
I added a new commit to this pull request to give the flatpak the org.cinnamon.ScreenSaver permission.

(It took a bit to get a flatpak to build and to get it to use my local git repo.)

Core447:
> I added a new commit to this pull request to give the flatpak the org.cinnamon.ScreenSaver permission.

Great work! Thank you!

PR #205, opened 2024-08-17
Added Image Layering to ActionBase

G4PLS:
You can now Layer images on top of each other by calling `set_layered_images()`, in that method you wont be able to set size, valign or halign because those things need to be done for the ImageLayer itself.

By default the size is 1.0, valign/halign are 0.0. Meaning the Image keeps the original size and is centered on the key.
By changing the size or alignment the image will be scaled and aligned based on the very first image in the array.

The very first image will in general be used as a base image, meaning the size of that image is relevant


Core447:
What do you think about moving this from the ActionBase into the DeckController to have a more standardized approach? We could then move away from KeyImage and KeyVideo to KeyLayer which can hold images, videos, text etc

G4PLS:
That would be nice but before we do that I would try getting layering videos working. Afterwards we can move it into the DeckController and rename ImageLayer to MediaLayer instead of KeyLayer, because MediaLayer just seems like a more appropriate name

I would also love to be able to use animated svgs so thats also someting im currently looking into

I also added the get_asset_path method to the ActionBase because always having to use the boilerplate `os.path.join(self.plugin_base.PATH, ......)` just got very annoying

G4PLS:
@Core447 so... animated svgs to a gif wont be possible. There is no library for it and doing it from scratch seems to big of a task so I dropped that

Wont be Transforming KeyVideo. I dont really understand KeyVideo so I wont really touch it. Even after reading the Code for it (And the needed classes for it), at least for now, its to big (For me) to implement

Core447:
> @Core447 so... animated svgs to a gif wont be possible. There is no library for it and doing it from scratch seems to big of a task so I dropped that

Ok, gifs should be sufficient anyways

> Wont be Transforming KeyVideo. I dont really understand KeyVideo so I wont really touch it. Even after reading the Code for it (And the needed classes for it), at least for now, its to big (For me) to implement

Ok, I can have a look at it when I'm back (and if you want to)

G4PLS:
Yeah gifs are sufficient but im currently working a lot with SVGs so it would have been nice if there is an easy way

Maybe I can get around to do it, but when you get to it before me thats completely fine. Im currently more active in the OBS Plugin and Im currently also creating a full guide to Plugins/Plugin Backends so im already quite busy

Core447:
> What do you think about moving this from the ActionBase into the DeckController to have a more standardized approach? We could then move away from KeyImage and KeyVideo to KeyLayer which can hold images, videos, text etc

Regarding this what do you think about moving all the layer stuff into a dedicated class with a method like `get_final_image`? This way devs can create their images with layers and show them using `set_media(image=self.layer.get_final_image())`.

G4PLS:
The current implementation has it so you have a static method that converts multiple layers into one image

The way you want to do it would mean we swap from a MediaLayer which only contains the information of one layer to a more general class maybe called Media, which contains an array of MediaLayers, that way we could make it so the Layers can be moved around and the overall Image can be moved around

This would also enable the `.get_final_image()` method

What are your thoughts on that @Core447

Core447:
That sounds good. I was thinking that in the long term we could move most internal image operations to the layers as well making a lot of the code easier to understand and maintain. Because of this something that can basically be a "drop-in" replacement (where we just have to use `get_final_image()` before we apply the image somewhere) would be very nice.

G4PLS:
Alright @Core447 this should do it. Im not able to test the change atm because I dont have time right now.
You are free to test the changes

It will also be very interesting how we get Videos working (With layering)

G4PLS:
@Core447 before merging feature should be tested by someone that didnt do any development on this feature
So either you or someone else should take a look

Core447:
Are the changes alright? If yes, I'm gonna merge

PR #208, opened 2024-08-20
Switch to xdg compliant folder structure

### LINKED ISSUE #207: Adhere to XDG specifications
**Describe the bug**
The apps puts all their data and config in `$HOME/.var`. This does no follow any specification, no other program uses such a directory.
Unfortunately this leads to a cluttered home directory and breaks setups that expect programs to put their data in a unified place.

**Expected Behaviour**
Ideally this program would follow the [XDG base directory spec](https://specifications.freedesktop.org/basedir-spec/0.8/).
This would mean:
- configuration such as the pages.json and static config should be put in `${XDG_CONFIG_HOME:-$HOME/.config}/streamcontroller`
- non configuration data such as plugin files should be put in `${XDG_DATA_HOME:-$HOME/.local/share}/streamcontroller`
- cached assets should be put in `${XDG_CACHE_HOME:-$HOME/.cache}/streamcontroller`

PatrickDaG:
This will close #207.
It's not yet extensively tested but just a suggestion about how I think something like this should look.

A migration is still missing because I don't know how this would preferably look, I think the main option we have are either checking if the old folder exists and then using the old folder structure, or shipping some kind of migration script, that moves the folders. First option is safer for users as we don't have to move files around but more annoying to implement as we basically have to add an if else anywhere a file is accessed. Second options allows us to migrate the whole codebase to the new layout but we need to be very careful implementing migrations as to not break anything on users machines.

While the old main folder `.var/app` is the default [Flatpak](https://docs.flatpak.org/en/latest/conventions.html) folder it does not sound like it is intended for apps to access  the path directly but follow xdg specs and use the environment variables, which will always be set by Flatpak.


maehne:
I fully agree that switching to an XDG compliant folder structure is a good thing. The [Flatpak](https://docs.flatpak.org/en/latest/conventions.html) documentation makes it also clear that it internally ensures the mapping to `.var/app`. Regarding your PR, I wonder, why you use the environment variables instead of the dedicated [GLib API](https://amolenaar.pages.gitlab.gnome.org/pygobject-docs/GLib-2.0/functions.html#gi.repository.GLib.get_user_config_dir)? `gi` is anyway already imported in other places.

PatrickDaG:
No particular reason. Just that I haven't worked with GLib before and don't know their API so it was easier for me to rely on the env which I know how to do. I can switch to the API if preferred.

Core447:
Great work, thank you! I have two ideas how I would do the migration tool:
1. Use the already existing migrator class handle this with the next app update
2. Save an invisible status file somewhere when the files have been migrated

We just have to keep in mind that it's possible to use a completely separate data path. If this is the case, we ofc don't want to move any files around.

Core447:
Gonna merge it into `xdg` and will push it to main when we have a migration tool

G4PLS:
Hows it looking with the migration tool?

PR #216, opened 2024-08-26
Enable creating flatpaks from local repo

jfbauer432:
After doing everything manually once to create a flatpak from my local git repo, I modified the install.sh script to allow creating flatpaks from other then the official location.  One future change that could be handy is to be able to specify any commit id to use.  But that would have complicated the editing of the .yml file and I really didn't need it for what I was doing.

Hope this is helpful.

-----------------------

Changed the flatpak/install.sh to it can be used to create a flatpak from a source other then the official repo on github. Also optionally create a flatpak bundle (can be used to test flatpak on another system).

Added a few options:

```
  -h --help             Show this message
  --repo=path           Path to StreamController repo (must be local)
                        use 'current' for git repo in current pwd
  --branch=branch       Name of branch in --repo to use
                        Ignored if --repo is not specified
  --make-bundle         Create a flatpak bundle so you can try
                        it on another system
  --yes                 Answer yes to all questions
```

Simplified the question answering code (needed after adding the --yes option)

### DIFF
--- a/flatpak/install.sh
+++ b/flatpak/install.sh
@@ -1,9 +1,125 @@
 #!/bin/bash
+#
+# Create a flatpak of StreamController and optionally a flatpak bundle
+#
 
 # Function to check if a command exists
 command_exists() {
     command -v "$1" >/dev/null 2>&1
 }
+
+usage() {
+    echo "Usage: $0 [options]"
+    echo "Options:"
+    echo "  -h --help             Show this message"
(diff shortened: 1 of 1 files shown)


Core447:
Thank you!, However, I'll need some time to go through this - I hope you can understand.

jfbauer432:
Take your time.  I realize it is a fairly big change to that file.  Feel free to use all, some, or none of it.

Issue #214, opened 2024-08-24
BUG: Autostart (non-flatpack!) and start in background do not work


3urobeat:
**Describe the bug**
Enabling autostart in the StreamController settings has no effect, the application needs to be started manually.  
This is caused by the software assuming it was installed as a flatpak, which it was not. Therefore the flatpak run command which was added to the autostart list has no effect.

~~I then tried to just add `streamcontroller -b` to my autostart list manually, but jokes on me, `-b` has no effect and the GUI always pops up.~~  
The only error messages in the log (nothing background related as it seems to me) are:
```
ERROR    | __main__:quit_running:205 - org.freedesktop.DBus.Error.ServiceUnknown: The name is not activatable
ERROR    | src.backend.GnomeExtensions:connect_dbus:31 - Failed to connect to D-Bus: org.freedesktop.DBus.Error.ServiceUnknown: The name is not activatable
ERROR    | src.backend.WindowGrabber.WindowGrabber:init_integration:59 - Unsupported environment: kde with server: wayland for window grabber.
```

~~On a side note, could you please add the flags `-h` & `--help` which output information about the binary and its available flags instead of starting the application? It's a very irritating behavior.~~

I'm on EndeavourOS (Arch) with KDE Wayland and StreamController was installed through the AUR.

&nbsp;

**To Reproduce**
1. Install something Arch based
2. Install StreamController as something non-flatpack (e.g. AUR: `yay -S streamcontroller`)
3. Enable autostart in the StreamController settings
4. Check the autostart list (e.g. in the KDE Systemsettings application) and observe that a flatpack command was added

~~Repeat 1 & 2, then:~~

3. ~~Run `streamcontroller -b` in your terminal~~
4. ~~Observe the application starting in the foreground~~

&nbsp;

**Expected behavior**
- Add a "normal" startup command to the autostart list
- ~~Respect the background flag~~
- ~~Output help instead of starting when running with `-h` or `--help`~~

&nbsp;

Thank you!

PS: Copy pasting buttons worked for me once in the very first session and have never worked since. CTRL+C & V do nothing anymore and right click copy makes the selection jump to the next button on the right. Is this something you know of?  
Also, the OBS plugins just never manages to connect and throws the error `AttributeError: 'NoneType' object has no attribute 'get_connected'`

&nbsp;

[...]

PR #218, opened 2024-08-29
fix: Add desktop entry for non-flatpak installations

3urobeat:
This PR attempts to fix autostart not working on non-flatpak installations.  
It does so by adding a `-native.desktop` file (slightly misleading in the flatpak subdirectory but I didn't find any other fitting place) which is used instead of the "normal" desktop entry when `is_flatpak() == False`.

I have tested this change on my non-flatpak Arch-based installation and on a flatpak Debian-based VM. Both chose the appropiate entry.

This fixes issue #214.  
Please let me know if I have made any oversight.

### DIFF
--- a/autostart.py
+++ b/autostart.py
@@ -32,7 +32,7 @@ def setup_autostart(enable: bool = True):
             setup_autostart_flatpak(True)
 
         else:
-            setup_autostart_desktop_entry(True)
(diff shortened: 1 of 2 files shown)


Core447:
Great work, I'll try this as soon as possible

Core447:
Thank you!

PR #219, opened 2024-08-30
fix: Fix store thread crash on plugin uninstall & broken plugin update indicator

3urobeat:
This PR fixes a crash of the store thread when trying to check for actions which need to be removed when uninstalling a plugin.

The `isInstance` check failed because it tried to access an element out of bounds.  

This has been tested on a non-flatpak installation. The code now correctly removes actions for the uninstalled plugin and then removes the plugin directory.

I have not yet tested this for a flatpak installation (need to go real quick, will do later).

Edit: Same behavior on flatpak, fix confirmed to be working. However the actions of removed plugins are only removed after reoping the store another time, on a non-flatpak installation this is happening instantly. Not exactly sure why.

### DIFF
--- a/src/backend/PageManagement/Page.py
+++ b/src/backend/PageManagement/Page.py
@@ -316,7 +316,7 @@ class Page:
         for type in list(self.action_objects.keys()):
             for key in list(self.action_objects[type].keys()):
                 for state in list(self.action_objects[type][key].keys()):
-                    for index in list(self.action_objects[type][key].keys()):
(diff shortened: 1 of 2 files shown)


3urobeat:
Added a commit to fix missing FETCH_HEAD when cloning plugins using git. This caused the update indicator to be broken, as well as an error message appearing in the log when opening the store window.

I did not want to open another PR for a small change when this already affects the plugin/store as well.

Core447:
Great work, thank you!

PR #221, opened 2024-08-31
fix: Fix plugin installation scripts missing dependencies in venv

3urobeat:
This PR fixes a bug where plugin `__install__.py` scripts always failed to execute when depending on modules not installed globally.

This was caused by the store using the global python binary instead of the one it was launched with itself.

Because of this bug, the OBS plugin for example was entirely unusable outside of global/flatpak installations.

This fixes the issue (can be closed after merging): https://github.com/StreamController/OBSPlugin/issues/10  
Issue was also mentioned in (this issue depends on other PRs): https://github.com/StreamController/StreamController/issues/214


Core447:
Great work, thanks!

PR #226, opened 2024-09-09
don't crash when moving a button that doesn't have functionality (e.g. image only)

violet4:
i just found StreamController a few minutes ago! awesome!

one of the first things i did was add an image to a button and then, without adding any functionality to the button, drag the "button" to another physical button. an exception is raised and the button fails to move as expected.

this is a naïve fix without really trying to understand the code, but it certainly prevented an error for me and enabled the functionless image button to be moved freely.

feel free to tell me it's awful and ideally give some pointers on the preferred way of handling this error and i'd be happy to put work in to make it better.

### DIFF
--- a/globals.py
+++ b/globals.py
@@ -15,6 +15,8 @@ argparser.add_argument("--skip-load-hardware-decks", help="Skips initilization/u
 argparser.add_argument("--close-running", help="Close running", action="store_true")
 argparser.add_argument("--data", help="Data path", type=str)
 argparser.add_argument("--change-page", action="append", nargs=2, help="Change the page for a device", metavar=("SERIAL_NUMBER", "PAGE_NAME"))
+argparser.add_argument("--list", default=False, action="store_true", help="List all the detected Stream Decks and exit")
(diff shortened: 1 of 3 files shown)


Core447:
> i just found StreamController a few minutes ago! awesome!

Thank you!

Is it ok if I just apply this directly?:
image
The problem with one try is that in case the first line raises an exception the second one isn't event attempted. This isn't a problem on the buttons but on the SD+ when dragging between dials, buttons and the screen.

You can also create a new pr for it if you want to. And please use a separate branch in your fork for every PR to avoid duplicated commits (like in this case where the "ability to run specified..." is in this PR as well)

Core447:
@violet4 What do you think?

violet4:
@Core447 yeah, go ahead and apply it directly! thank you for looking into it! 😊

ahh, thanks for the note about the separate branches. tl;dr that wasn't supposed to happen but now i understand PR/branch relationship better and won't do it again 😂️

Core447:
> ahh, thanks for the note about the separate branches. tl;dr that wasn't supposed to happen but now i understand PR/branch relationship better and won't do it again 😂️

No problem, it happened to me as well.

The changes are now available with 6f1694d1e1c943608b31878645397b515a5cc72f.

PR #227, opened 2024-09-09
ability to only run specified Stream Decks instead of all attached decks

violet4:
this is achieved by adding `--list` and `--paths` command line arguments.

* `--list` lists all the attached Stream Deck devices with some basic information, including their "paths", and exits
* `--paths` takes 1 or more paths (found via `--list`) and filters out paths not specified before initializing the devices

if anyone cares about this change enough to want to merge it, but it's terrible or doesn't match the project requirements, please point me in the correct direction and i'd be happy to improve it and make it acceptable. however, at the moment, this makes it so i can use this software with one of my stream decks and my own program with my other stream deck simultaneously.

### DIFF
--- a/globals.py
+++ b/globals.py
@@ -15,6 +15,8 @@ argparser.add_argument("--skip-load-hardware-decks", help="Skips initilization/u
 argparser.add_argument("--close-running", help="Close running", action="store_true")
 argparser.add_argument("--data", help="Data path", type=str)
 argparser.add_argument("--change-page", action="append", nargs=2, help="Change the page for a device", metavar=("SERIAL_NUMBER", "PAGE_NAME"))
+argparser.add_argument("--list", default=False, action="store_true", help="List all the detected Stream Decks and exit")
+argparser.add_argument("--paths", default=[], nargs='+', help="Open only the Stream Decks with the given paths (see paths using --list)")
 argparser.add_argument("app_args", nargs="*")
 
 VAR_APP_PATH = os.path.join(os.path.expanduser("~"), ".var", "app", "com.core447.StreamController")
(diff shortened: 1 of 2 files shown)


Core447:
Nice work! In the future we can maybe even add an option to the settings for this. The only problem I'm encountering is that if I instance is already running and `--list` is used it shows the window of the running instance and doesn't print the available decks.

violet4:
tl;dr go ahead and close this PR when you see my message and i'll eventually submit a better one that handles more cases.

also my PR doesn't work so great if the externally managed stream deck is connected while StreamController is already running, since StreamController seems to watch new stream deck connections and attach to them automatically.

so clearly, this naïve/lazy PR isn't going to work for most use cases and i'm gonna have to put some more work into making it more robust and generally acceptable.

i'm not in any rush to make this super fully integrated and handle all cases, but i'm thinking this PR should be closed and i can use my own branch for a while over the course of a period of time and make small fixes until it's been a while since i've seen any issues. i can continuously rebase from this repo so it stays in sync and eventually make a PR when i've fixed all (currently unhandled) cases i've encountered (plus the one you mentioned).

i'll let you see this message, feel free to close the PR and i'll (slowly) work on improving it on my branch. at the moment my main concern is ensuring compatibility with my own awful stream deck implementation that i depend upon daily until i get the functionality moved over to StreamController.

((for example, i'll have to learn how the plugin system works so i can port my cute little stream deck calculator app.))

thank you for your time and patience!

violet4:
also i created a ticket for myself to keep track of specific needed fixes https://github.com/violet4/StreamController/issues/2

Core447:
> port my cute little stream deck calculator app

Sounds nice!

@violet4 I'll convert this PR to a draft so that you don't have to create a new one (which you can ofc also do if you want).

PR #232, opened 2024-09-16
BUGFIX: Active window not found

### LINKED ISSUE #229: BUG Page autoswitching does not work in Ubuntu XFCE4 X11
**Describe the bug**
Neither on master nor on 1.5.0-beta6 page autoswitching works when used on Ubuntu (24.04.1), XFCE4 (4.18), X11 (xserver-xorg - 1:7.7+23ubuntu3)

**To Reproduce**
Create a page for autoswitching (in my case for MS teams webapp)
Add matcher on 'teams'
Check that single window is matched

Switch to MS Teams web app

Page is not switched

**Expected behavior**
Page is automatically switched

**Screenshots**
Attached sanitized page
[Teams.json](https://github.com/user-attachments/files/16993324/Teams.json)

zeridon:
In some cases xprop returns more window id's than just the active one.

Adopt `get_all_windows` approach to iterating over all the window id's and get the first one that is not `0x0`

Fixes https://github.com/StreamController/StreamController/issues/229

### DIFF
--- a/src/backend/WindowGrabber/Integrations/X11.py
+++ b/src/backend/WindowGrabber/Integrations/X11.py
@@ -96,6 +96,8 @@ class X11(Integration):
                 return
             stdout, stderr = root.communicate()
             window_ids = stdout.decode().strip().split("#")[1].strip().split(", ")
+            if len(window_ids) == 0:
+                return
             for window_id in window_ids:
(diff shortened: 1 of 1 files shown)


Core447:
Thank you!

PR #234, opened 2024-09-19
Fix errors in info panel for wallpapers

zeridon:
When opening the info page for a wallpaper pack only errors are displayed [example](https://github.com/user-attachments/assets/c0011aef-9341-40ea-a0a6-cd4b9cf7dd16).

This PR fixes the issue


Core447:
Thank you!

PR #236, opened 2024-09-21
Fix loading of locales in certain cases

### LINKED ISSUE #225: BUG: StreamController fails to launch
**Describe the bug**
StreamController fails to launch out of the blue with the error message `2024-09-07 13:48:02.642 | ERROR    | __main__:quit_running:204 - org.freedesktop.DBus.Error.ServiceUnknown: The name com.core447.StreamController was not provided by any .service files

**To Reproduce**
1. Something happens to the system (haunted?)
2. Launch StreamController from the terminal to see debug messages with `flatpak run com.core447.StreamController`
3. It throws the error message2024-09-07 13:48:02.642 | ERROR    | __main__:quit_running:204 - org.freedesktop.DBus.Error.ServiceUnknown: The name com.core447.StreamController was not provided by any .service files

**Expected behavior**
StreamController launches

**Screenshots**
Screenshot from 2024-09-07 13-52-12

**Additional context**
StreamController did not start at startup like it was supposed to either, and launching via Shortcuts has the same effect.
System: Ubuntu 22.04
Reinstall of StreamController was attempted, but produced the same result

etienne02:
Change the locales path for loading to an absolute path (based on the main.py file) in order to allow launching the StreamController from any path. Currently we can only loads it with workdir being the source root of the project otherwise locales won’t be found and project will refuse to start.

May fix #225

### DIFF
--- a/main.py
+++ b/main.py
@@ -60,6 +60,8 @@ from src.backend.Migration.Migrators.Migrator_1_5_0_beta_5 import Migrator_1_5_0
 # Import globals
 import globals as gl
 
+main_path = os.path.abspath(os.path.dirname(__file__))
+
 def write_logs(record):
     gl.logs.append(record)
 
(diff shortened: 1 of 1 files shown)


Core447:
Great catch, thanks for fixing!

PR #238, opened 2024-09-21
Fix threads

etienne02:
Fixes colliding _close in the thread (that collide with Thread._close()) and make the application not closing properly.
Fix some deprecated method usages in threads (getName(), setName(), ... are replaced by properties)

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -147,7 +147,7 @@ class App(Adw.Application):
         # Force quit if normal quit is not possible
         timer = threading.Timer(6, self.force_quit)
         timer.name = "force_quit_timer"
-        timer.setDaemon(True)
+        timer.daemon = True
         timer.start()
 
         for ctrl in gl.deck_manager.deck_controller:
--- a/src/backend/DeckManagement/DeckController.py
+++ b/src/backend/DeckManagement/DeckController.py
@@ -304,6 +304,7 @@ class MediaPlayerThread(threading.Thread):
             self.touchscreen_task.run()
             del self.touchscreen_task
             self.touchscreen_task = None
+
     def check_connection(self):
         try:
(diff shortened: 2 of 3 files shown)


Core447:
Thank you! Haven't had the time yet to really go through everything, but what I noticed so far is that you changed the lines with: `Set stop flat - otherwise remove_controller will wait until this task is done, which it never will because it waiuts` (typo needs to be fixed) to: `.stop()`. However, using `.stop()` will cause an infinite loop because the task of the thread is waiting until the task is completed but the task only completes when the thread ends.

etienne02:
> Thank you! Haven't had the time yet to really go through everything, but what I noticed so far is that you changed the lines with: `Set stop flat - otherwise remove_controller will wait until this task is done, which it never will because it waiuts` (typo needs to be fixed) to: `.stop()`. However, using `.stop()` will cause an infinite loop because the task of the thread is waiting until the task is completed but the task only completes when the thread ends.

For what I tested there is no infinite loop. The stop method setup set the stop condition to true and then join it (wait for it to stop), because this is a loop that doesn’t hang (waiting for something), but regularly do a loop step, it’s fine. I can add a timeout to the join to ensure that the loop can’t wait forever.

Core447:
I'm pretty sure that there was a reason I did it via the flag and because I don't see a reason to change this to `.stop()` I think we can just let it as it is. Gonna take a look at the other changes soon.

etienne02:
> I did it via t

In the current code base status, there is no risk of dead lock (thread waiting itself to stop) because, it’s closing another thread.
I find it more clean to have a call to stop. and if we want we can remove the join call to the stop method and do a join call after a stop call where it is necessary.

PR #243, opened 2024-09-22
KDE Application Switching: Wayland

Electro707:
This PR adds native KDE support for window/app switching, which is mostly because of Wayland. But it should also work on X11 with KDE

TODO:
- [x] Add a closing handler to stop the KDE script
- [ ] Make sure works with Flatpack

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -162,6 +162,9 @@ cython_debug/
 # VSCode
 .vscode/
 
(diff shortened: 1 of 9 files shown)


Core447:
Hey, sorry for the huge delay. I'm gonna close this in favor of #308 as it works with Flatpak

PR #258, opened 2024-10-15
Move to Gnome 47 runtime

wanderboessenkool:
Apart from moving to the latest and greatest Gnome flatpak runtime this also solves the `follow_symlinks` argument unknown in pathlib, since the 47 runtime uses Python 3.12, whereas the older 46 runtime has Python 3.11 where pathlib does not yet know about that argument.


Core447:
Thank you!

PR #263, opened 2024-10-22
Fixed get_asset_path in ActionBase

G4PLS:
With `subdirs: list[str] = []` the possibility was that modifying the subdirs list could persist. Using `None` is overall the better option when wanting a List as a param.


Core447:
Thank you!

PR #266, opened 2024-10-26
[FEAT]: Search Combo Row

G4PLS:
Added a Combo Row which has the ability to search

This is quite useful when you have lots of items in your comborow/dropdown. The current implementation is not perfect but it works quite well


Core447:
Great, thank you! I'm gonna update the page selector to use this.

G4PLS:
I noticed some smaller issues and I will try to fix those

Currently the search has to 100% match, you cant click on an item to select it (Dont know why yet)

PR #267, opened 2024-10-28
Fix(SearchComboRow): Changed instances of ComboRowSearchItem to SearchComboRowItem

G4PLS:
When changing the names of the classes to the final version I forgot to change the previous naming "ComboRowSearchItem" to "SearchComboRowItem", resulting in errors when using the SearchComboRow


G4PLS:
I have now also added the new feature to be able to select an item by index

PR #269, opened 2024-10-31
Add Stream Deck NEO to udev.rules

sifmelcara:
Hi, nixpkgs StreamController package maintainer here. We packaged `1.5.0-beta7`, although changlog says Stream Deck Neo support is added, udev rule seems to be missing (we install this `udev.rules` file into user's system)

Thanks for the awesome project!


Core447:
Alright, thank you!

PR #271, opened 2024-11-09
Feat(Locales): Updated PluginBase and LocaleManager for minor improvements

G4PLS:
- Added two args to __init__ to define the LegacyLocaleManager
  - use_legacy_locale: Defines what LocaleManager should be used. True = Legacy | False = Modern
  - legacy_dir: Made it possible to change the directory of the Locales for the LegacyLocaleManager

- Added a return to the load_csv in the LocaleManager so that the Plugin is still able to load even when the locales are missing

### DIFF
--- a/locales/LocaleManager.py
+++ b/locales/LocaleManager.py
@@ -16,6 +16,7 @@ along with this program. If not, see <https://www.gnu.org/licenses/>.
 import os
 import csv
 import locale
+from loguru import logger as log
 
 class LocaleManager:
     def __init__(self, csv_path: str) -> None:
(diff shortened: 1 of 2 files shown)


Core447:
Thank you!

We might need to set the legacy locale manager as the default, for backward compatibility (because all plugins use the legacy atm).
@G4PLS What do you find more intuitive, the json files of the legacy manager, or the csv of the new one?

G4PLS:
I like the csv files a lot more, I personally will use the new one from now on

But yeah I will change the default from False to True

PR #274, opened 2024-11-09
Feat(EventAssigner): Reworked and added the ability of Action Events

G4PLS:
This should make the Event Assigner better by only showing events that are actually in an Action
This will rework some of the actual Event Assigner and the ActionBase to accommodate those changed

### DIFF
--- a/src/backend/PageManagement/Page.py
+++ b/src/backend/PageManagement/Page.py
@@ -447,10 +447,20 @@ class Page:
 
         if action_object is None:
             raise ValueError("Could not find action object")
+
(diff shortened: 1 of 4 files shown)


Core447:
Looks good to me, one problem I noticed, though is with localization: event_callback gets passed the same value as shown in the ui, e.g. "Down key", but we might want to show "Taste runter" in the ui but keep sending "Down key" (or some constant key) to event_callback.

G4PLS:
Yeah thats true
But tbh im still not very happy with the system, it feels quite annoying to me personally, but maybe thats only because i developed it

PR #275, opened 2024-11-10
Fix(ActionConfigurator): UI not getting hidden properly

G4PLS:
The Rows or Custom Config were not hidden properly when the methods were not present in the Action.
Changed it so we compare the actual method and hide if the method is the same, meaning it was not overridden.
Also made the 2nd separator hide as this would not look nice without it

### DIFF
--- a/src/windows/mainWindow/elements/Sidebar/elements/ActionConfigurator.py
+++ b/src/windows/mainWindow/elements/Sidebar/elements/ActionConfigurator.py
@@ -157,10 +157,13 @@ class ConfigGroup(Adw.PreferencesGroup):
         pass
 
     def load_for_action(self, action: ActionBase):
-        if not hasattr(action, "get_config_rows"):
+        base_method = getattr(ActionBase, "get_config_rows", None)
+        action_method = getattr(action.__class__, "get_config_rows", None)
+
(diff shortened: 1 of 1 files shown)


Core447:
Unless, I'm missing something, the rows already get hidden by [L206](https://github.com/StreamController/StreamController/blob/2b5e3284ae272750ed522643fcdf5ea7dc1a6444/src/windows/mainWindow/elements/Sidebar/elements/ActionConfigurator.py#L206-L207) and [L165](https://github.com/StreamController/StreamController/blob/2b5e3284ae272750ed522643fcdf5ea7dc1a6444/src/windows/mainWindow/elements/Sidebar/elements/ActionConfigurator.py#L165-L166). And the problem with your approach of comparing the methods is that if a dev overrides the default methods, but still returns None it won't hide the groups because the method has been overridden.

G4PLS:
While developing I noticed that it didnt get hidden
Or maybe it was the separator that just got displayed double, so at least a check should get added that hides the 2nd separator

G4PLS:
This should now be a better fix
Noticed the rows always return an empty list so the check for None was wrong

This should now fix it, I also removed your hasattr check as this would never be evaluated correctly

PR #287, opened 2024-11-23
Using --change-page argument no longer pops up main window

dixonte:
A simple change to the startup order, allows --change-page to affect running instances without opening the main interface.

### DIFF
--- a/main.py
+++ b/main.py
@@ -220,7 +220,7 @@ def quit_running():
 
 def make_api_calls():
     if not gl.argparser.parse_args().change_page:
-        return
+        return False
     
     session_bus = dbus.SessionBus()
     obj: dbus.BusObject = None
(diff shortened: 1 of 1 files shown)


Core447:
I think the current behavior might make more sense. This way all api calls are api calls and no longer arguments to control the starting window. The default pages can still be configured in the ui, so I don't think there's a real use case for `--change-page` that keeps opening the window.

If you think differently about this, let me know.

Core447:
You can also have a look at the `api-calls` branch, where I'm in the progress of re-doing a lot of this

dixonte:
So my purpose with this PR was to be able to change the current page displayed on the Deck without popping up the frontend UI, regardless of if the frontend is running already or not. This is not possible in the branch this PR was based upon.

If the api-calls branch has a feature to do this, then I would agree this PR is unnecessary.

Core447:
Ok, then I think I misunderstood you. I just tested the `--chage-page` argument, and now I see what you mean.

Gonna merge it. Thank you!

PR #289, opened 2024-11-26
Improve window grabber support for swaywm

jgoett154:
The initial swaywm window grabber implementation did not pick up most of my windows. Unsure if it's a swaywm version difference or just due to me using tabbed and stacked layouts but this recursively walks the tree and picks up all windows for me now. Been testing for about a day and a half swapping between several windows automatically and it seems to be working!


Core447:
Alright, thank you!

PR #295, opened 2024-12-08
Feat(ActionGroup): Added Action Groups

G4PLS:
- Added groups to the PluginBase that can be added with `add_action_group`
- Added a new Expander to the ActionChooser called ActionGroupExpander
- Added some Styling to the UI elements

This is the current UI for the Groups. When adding any styling to the nested expanders they go this black so I tried to match the Group Items to that styling. Using nested expanders also break the hover of normal items so I added that manually with more styling
image


G4PLS:
This is how the new styling looks, this can easily be changed in style.css
image

G4PLS:
Latest styling now looks like this:
image

Core447:
Nice work! The only problem I noticed is that the arrow on the right hand side always shows the "open state". Idk how if that's really fixable, as expanders inside expanders behave a bit strange.

G4PLS:
> Nice work! The only problem I noticed is that the arrow on the right hand side always shows the "open state". Idk how if that's really fixable, as expanders inside expanders behave a bit strange.

Jup thats something to take a look at
But i personally dont find it that bad if its not fixable (It would be annoying tho)

Searching is also something that needs to be reworked. Currently you get errors because the Expander dont contain a label

G4PLS:
@Core447 do you maybe know how to remove the arrow?
Then I could implement that behaviour manually

Core447:
@G4PLS [a61a91e](https://github.com/StreamController/StreamController/pull/295/commits/a61a91e85fa4896410526d7f8b02105be0b3a8c1) fixes this

Core447:
Tbh I don't really like the new position of the store button, as it's not directly clear what it does. Also with the old one at the bottom, if a user searches through the actions from the top to the bottom and doesn't find anything, he'll see the open store button right away when he needs it.

G4PLS:
Ah yeah that makes sense
I personally liked the new one more because its quite clear (The Icon im using is imo very well known)

But we can ofc move it back to the bottom

G4PLS:
@Core447 we need to add the Function back that disables the Action if its unsupported/untested
Currently everything gets shown normally. I dont know how you did this so would it be possible for you to take a look or can you tell me how you determined that?

G4PLS:
We also need a nicer way to style it, currently its visually now that easy to seperate between Actions and Groups
image

G4PLS:
ActionHolderGroups now get defined like this in main:
```py
self.group_1 = ActionHolderGroup("C-D", action_holders=[self.e, self.f], hide_on_group_support=False, group_support={Input.Key: ActionInputSupport.SUPPORTED})
self.add_action_holder_group(self.group_1)
```

hide_on_group_support and group_support is optional

Core447:
@G4PLS I changed it so that the groups keep getting shown, even if no action is compatible, so that users can still see what is available. If you're ok with the current state, I'm gonna merge it

G4PLS:
@Core447 I had this code in place so devs could determine if they want the actions to be shown if the group is available for the set action.

By default every group should be shown even when its not supported

Core447:
Ok, gonna merge it now. We can still change this later on

G4PLS:
Alright!

I will still explain my thought process for the code I had in place:
When you are able to Hide groups that are not compatible you can create multiple groups with the same name that are intended for the different Input methods.

So multiple groups called recording but every group is for a single input method

PR #300, opened 2024-12-17
Update udev.rules

DELTAJoSch:
I got a permission error for /dev/uinput on PoPOS!. I have fixed this by adding an additional rule to the udev-rules file which allows access to the device. After a reboot, my installation now works.


Core447:
Did you get this error when running the app itself, or when using the OS plugin? Because the OS plugin shows a message when the input rule is missing with a link to a tutorial.

Core447:
Gonna close this, as this udev rule is only needed for the OS plugin, which clearly states that this rule is necessary and also provides a link to instructions

PR #306, opened 2024-12-29
Feat(Migration): Reworked Migration System

G4PLS:
The whole migration system got reworked and is now ready to be merged!
The migration System is now very Scalable and versatile

- [ ] Implement missing Migration from earlier versions

Because I didnt pay attention the Migration also includes the PluginSettings implementation


Core447:
- I've noticed that you pass in `gl.app_version` as the from_version in `Settings_1_5_0b8.py` ([here](https://github.com/G4PLS/StreamController/blob/f5d76c8d9823114678c536a72e81fe31b90bc14d/src/backend/Migration/Migrators/Settings_1_5_0_b8.py#L19)) which is what the version migrator expects, but the current handling of `from_version` doesn't make sense to me, as the app that should be migrated to is the one currently running. So if we have a migrator from version 1 to version 2 then the check will have version 2 as the `from_version` and the `to_version` which doesn't make sense. Also we have to keep in mind that users might skip app versions. Therefore we have to chronologically run the migration for each app version. The current implementation is doing this [here](https://github.com/StreamController/StreamController/blob/22190cd334af3180c6779e7566b4ed27ecd025d8/src/backend/Migration/MigrationManager.py#L28).
- I would suggest changing the `dependent_migrator` to `dependent_migrators` (a list) - This could also solve the problem above, by (maybe automatically) setting the migrators to old app versions as dependent
- We also have to set `self.migration_ran` true when the condition has not been met.
- As you already mentioned we need to re-implement the current migrators (at least the recent ones)
- And before the final merge we have to test this extensively, as an error in this system may block the app from launching after an update.

G4PLS:
1. Youre right, the VersionCondition needs to be setup in a better way, I will think about what would make more sense

2. The dependant migrator is basically a migration chain. Example:
We have 5 migrator M1->M5 and we have this chain:
M1->M2->M3 & M4->M5
Meaning before M3 can run M2 runs, and before that M1 runs, The same with M4 and M5

3. `migration_ran` is for the purpose of the actual migration running. Meaning the whole migration process success.
This is set in the `MigrationBase` at the very end of `migrate()`

Core447:
> 2\. The dependant migrator is basically a migration chain. Example:
>         We have 5 migrator M1->M5 and we have this chain:
>         M1->M2->M3 & M4->M5
>         Meaning before M3 can run M2 runs, and before that M1 runs, The same with M4 and M5

ok

> 3\. `migration_ran` is for the purpose of the actual migration running. Meaning the whole migration process success.
>         This is set in the `MigrationBase` at the very end of `migrate()`

But if we have the chain M1->M2->M3 and for some reason the check for M1 returns false, but M2 and M3 still have things to to. Then M1 will stop the entire chain. For me "dependent" refers more to: "The dependent migrator has finished his work - no matter if work has been done or not"

Issue #256, opened 2024-10-11
Support KDE Plasma Wayland for automatic page switching


StelardActek:
**Describe the feature idea**
The automatic page switching feature is about the only thing I'm missing from Windows. I'm assuming it doesn't work because I'm not on Gnome or X11. Hopefully support for Plasma and/or Wayland can be added?

If not, is there any workaround?

**Additional context**
n/a

Core447:
> I'm assuming it doesn't work because I'm not on Gnome or X11

Yes. 

> Hopefully support for Plasma and/or Wayland can be added?

Yes, but I need to find a way to get the active window's title and class programmatically.

StelardActek:
Ah, I see. Cursory searching suggests this isn't simple. :|

Igorgro:
Maybe this can be helpful https://github.com/jinliu/kdotool. It has `getactivewindow` command

Core447:
Closing this because of the merge of #308

PR #308, opened 2025-01-04
Add KDE window grabber integration using kdotool

flifloo:
This PR is for the issue #256

The binary [kdotool](https://github.com/jinliu/kdotool) can get window information for KDE (Wayland and X11)

I tested this on Arch with KDE Wayland without Flatpak, I based the integration class on the X11 one.


Core447:
Great work - Thank you!

Issue #329, opened 2025-02-07
BUG: RPyC not completely configured


Kekemui:
**Describe the bug**
Both `Server`s and `Connection`s need to be configured in RPyC to allow access to non-primitive items across the wire. However, only the `Server`s are currently configured with permissions. This means objects, including instances of dataclasses, cannot be accessed across the wire.

**To Reproduce**
1. In a Python runtime (*not* Flatpak), add https://github.com/Kekemui/VeadoSC/tree/6a-nonfunctional as a custom plugin
2. Download the plugin from the store
3. See `Error initializing plugin <class 'plugins.gg_kekemui_veadosc.main.VeadoSC'>: cannot access 'get_observer_id'` in logs

**Expected behavior**
Plugin moves through initialization correctly.

**Augmented Logs**
[...]

PR #330, opened 2025-02-07
fix(rpyc)(#329): add config to fe->be connection

Kekemui:
* Adds config to frontend-to-backend connection with same semantics as the frontend server in PluginBase.
* A separate change (in streamcontroller-plugin-tools) is required to remediate an identical issue in BackendBase (though impact from that omission was not observed as part of this defect).
* This change is not intended to change data accessibility beyond what is semantically expected, and as such does not add further affordances like `set_attr` (which would allow property setters) or other useful RPyC permissions.

## Testing

* Pulled `https://github.com/Kekemui/VeadoSC`/`6a-nonfunctional` as a custom plugin and verified cross-process data access functioned as expected.
  * (Note there are typing issues that inhibit further functionality of the plugin [specifically w.r.t. netrefs obscuring event types]. Full functionality of the plugin is neither expected nor required for validation of this PR, just that the plugin spins up.)

### DIFF
--- a/src/backend/PluginManager/PluginBase.py
+++ b/src/backend/PluginManager/PluginBase.py
@@ -555,7 +555,7 @@ class PluginBase(rpyc.Service):
         Returns:
             None
         """
-        self.backend_connection = rpyc.connect("localhost", port)
+        self.backend_connection = rpyc.connect("localhost", port, config={"allow_public_attrs": True})
         self.backend = self.backend_connection.root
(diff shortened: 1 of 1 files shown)


Core447:
Thanks a lot for this fix!

Atm we're using the `dev` branch for development, so I just went ahead and manually applied your changed with 81dd61e7b6ad248c3f15b5cb2968c42c158e826d - I hope that's ok.

PR #336, opened 2025-02-16
Initial Code for TrayIcon core implementation

eumario:
Copied over trayicon.py from plugin to core file in src/backend/trayicon.py Renamed TrayIcon and Menu in trayicon.py to DBusTrayIcon and DBusMenu Updated src/tray.py with code to utilize functionality of DBusTrayIcon. Implemented menus for Show Window, Settings, Store, About, and Quit. Due to lateness in getting MainWindow created, added gl.tray_icon.initlaize() to end of build() function for MainWindow, so that we can be sure that HamburgerMenuButton has been created for their actions.


Core447:
Amazing! https://github.com/Core447/com.core447.StreamController/commit/9ecded7e53d787d529fddafd13cd789bd9a1db88 will add the needed dbus permission to the Flatpak

PR #344, opened 2025-02-27
[improvement] only return usb devices made by elgato

SilentSwordmaiden:
if a pc has a lot of usb devices attached the query for all of them was very slow.

we can filter it by only listing devices made by elgato and then check the (hopefully smaller) list for the actual products.

this sped up the initialization from a few mins to 2.5 seconds on my pc.


Core447:
Thank you!

PR #351, opened 2025-03-05
Added option to rotate deck

gensyn:
Please let me know what you think. I tested with a Mini and Original and it works the way I want it to.


Core447:
Looks good. I'm not home for the entire week, but will try it when I'm back.

gensyn:
I had to update the PR because `DeckController.set_rotation()` calls `self.update_all_inputs()` and for that to work the whole class had to be initialized.

gensyn:
Locally I have also added changes to show the key grid in the app in the correct orientation. Displaying the grid works fine now, but I can't interact with the grid because the keys of the page are still mapped in the old orientation and I can't quite figure out how to handle this. Maybe you want to give it a try, though I think for a first release of this feature it is not strictly necessary.

Core447:
> Maybe you want to give it a try, though I think for a first release of this feature it is not strictly necessary.

Alright, I can implement the last small fixes. You already did an amazing job!

Core447:
image

I just need to fix a few things, then it should be ready to get merged!

Core447:
I think we're ready to merge. Could you maybe also try and see if it works for you?

Core447:
@gensyn I think we're ready to merge. Could you maybe also try and see if it works for you?

gensyn:
Sure, might take a couple of days though.

Core447:
Ok, then I'll merge now but won't create a new release until you checked it out. If something is not working we can always fix it later.

PR #353, opened 2025-03-10
Refactor(GenerativeUI): Some smaller changes

G4PLS:
- Made some values private as they normally shouldnt be changed after setting, this just makes it clear
- Using a more up to date standard for defining the generic in the class
- Renamed some methods as `*_key()` doesnt make sense because were setting the value, not the key
- `on_change` now expects a method with the exact signature that is defined in the class, providing better type hinting


Core447:
Thank you!

PR #354, opened 2025-03-11
Feat(GenerativeUI): Added translation method

G4PLS:
This Implements a simple method that returns the translated text given a string.

To auto translate titles or subtitles a simple `get_translation` call can be made

In here we pass the title as the key and as the fallback, meaning it tries to auto translates the title and uses the title as a fallback if nothing was found. 
```py
self.widget: Adw.SwitchRow = Adw.SwitchRow(
    title=self.get_translation(title, title),
    subtitle=self.get_translation(subtitle, subtitle),
    active=default_value
)
```

### DIFF
--- a/GtkHelper/GenerativeUI/GenerativeUI.py
+++ b/GtkHelper/GenerativeUI/GenerativeUI.py
@@ -69,3 +69,9 @@ class GenerativeUI[T](ABC):
     def load_initial_ui_value(self):
         value = self.get_value()
         self.set_ui_value(value)
+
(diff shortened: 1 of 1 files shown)


Core447:
Isn't the new method not used at all atm? Cause `self.get_translation` is just calling the already existing method of the ActionBase. Or am I overlooking something?

G4PLS:
No that is correct, but having it makes getting translations one step easier

If this should get merged I will start on adding the actual Widgets which actually call the added method

PR #355, opened 2025-03-12
[FEAT]: Adding Generative UI Elements

G4PLS:
With this im adding the basic UI Elements that could be used for Generative UI.

It also adds a few improvements and fixes encountered while implementing said UI. Some Commits have some extra comments added aswell to see what my thought process was when adding the Change/Implementation

PR #358, opened 2025-03-13
Added option to set an SVG string as media

gensyn:
The HomeAssistantPlugin sets SVG icons as images on the keys and until now I have created the images inside the plugin. Since you built SVG support into StreamController, it felt like I could piggy back on your implementation by allowing not only paths to SVG files, but also respecting strings containing SVG data.
I already changed the implementation in my plugin and everything looks good.

### DIFF
--- a/src/backend/DeckManagement/HelperMethods.py
+++ b/src/backend/DeckManagement/HelperMethods.py
@@ -36,22 +36,23 @@ from autostart import is_flatpak
 import globals as gl
 
 
-def sha256(file_path):
+def sha256(text: str) -> str:
     """
-    Calculates the sha256 hash of a file.
+    Calculates the sha256 hash of a file or string.
 
     Args:
-        file_path (str): The path to the file.
+        text (str): The file path or string.
 
     Returns:
-        str: The sha256 hash of the file.
+        str: The sha256 hash of the file or string.
     """
     hash_sha256 = hashlib.sha256()
-    if not os.path.exists(file_path):
-        return
-    with open(file_path, "rb") as f:
-        for chunk in iter(lambda: f.read(4096), b""):
-            hash_sha256.update(chunk)
+    if os.path.exists(text):
+        with open(text, "rb") as f:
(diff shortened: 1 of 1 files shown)


gensyn:
~~Correction: Not everything looks good. Opacity is not working anymore. Let me look into that before possibly merging.~~

I'm stupid. It's working fine. The downside is, that for every opacity level a new file is written to the cache. I have a slider for opacity that I slided from 100 to 0 and now I have 101 new files in the cache - that's not very clean. Maybe images from SVG strings shouldn't be cached?

Another idea: On startup, delete cached files that were created - say - more than a week ago?

Core447:
> Another idea: On startup, delete cached files that were created - say - more than a week ago?

Yeah, that would work. I'll look into this.

gensyn:
I was kinda wondering why you are caching the images to begin with? In case the orginial gets (re)moved? I think caching should not be necessary for SVG-strings.

G4PLS:
I think that caching the Images in general for actions is probably not needed as they get added/created at runtime anyways. Especially now if you use the asset manager

For things like the Wallpaper Packs or the Icon Packs it makes sense

Core447:
I can't recall the exact reason, but I assume there were some problems going converting svgs into PIL objects in-memory. Nevertheless, 5529edaa9928881a348d2516dde55f5c452e87fd removes the caching and does everything in-memory

Core447:
Haven't tested the merge yet. Can you please check that everything is still as you wanted it to be?

gensyn:
Looking good from my end. Icons are shown, scaling and opacity work as expected, no files in the cache folder. Should be good to merge, thanks.

gensyn:
Idea: You could retrieve the old value before writing the new one and emit a signal yourself with both values.

Core447:
Great - Thanks for your work!

gensyn:
When will this change be rolled out to users?

Core447:
With the next release, but I can't say for sure when this will be, as I want to merge #351 and #355  first.

PR #365, opened 2025-03-23
Feat(ComboRow): Implemented differenciation between display and save value

G4PLS:
@Core447 @gensyn could you test if everything with that still works the same?


gensyn:
Yeah, but probably not before wednesday.

Issue #372, opened 2025-03-28
[Feature Request] Hyprland screen lock support


dennisrijsdijk:
**Describe the feature idea**
Add support for Hyprland's [wayland lock protocol](https://wayland.app/protocols/hyprland-lock-notify-v1) for the "lock decks when screen is locked" feature

**Additional context**
I have found a [library](https://github.com/grking/python-wayland) that can be used for this purpose, although it needs some slight modifications to add support for hyprland's protocols.

I'm willing to work on this feature if desired.

Core447:
> I'm willing to work on this feature if desired.

Sure, thank you!

Core447:
Added with #374

PR #374, opened 2025-03-30
feat: lock screen support for Hyprland

dennisrijsdijk:
Relevant Issue: #372 

This pull request implements a wayland socket communicator and hyprland screen lock support.

Testing:
Ensured app still launches under X11 and XWayland.
When launched without `WAYLAND_DISPLAY` (either X11 or XWayland), the `gl.wayland` global is not created.

To be resolved in draft:
New message format for lock screen settings message
- Seeing as there is now support for GNOME, KDE, Cinnamon and Hyprland, I feel like tacking on each supported DE/WM will end up causing a very long text entry if more are added in the future. Looking for feedback for the message format here

reqs.yaml and pypi-requirements.yaml
- I'm not sure if these should be updated within each PR, or if that's a packaging job.
- I can't seem to generate the pypi-requirements.yaml file. It looks distinct from reqs.yaml in the current upstream commit, and both files only include the command to generate reqs.yaml

### DIFF
--- a/main.py
+++ b/main.py
@@ -52,6 +52,7 @@ from src.Signals.SignalManager import SignalManager
 from src.backend.WindowGrabber.WindowGrabber import WindowGrabber
 from src.backend.GnomeExtensions import GnomeExtensions
 from src.backend.PermissionManagement.FlatpakPermissionManager import FlatpakPermissionManager
+from src.backend.Wayland.Wayland import Wayland
 from src.backend.LockScreenManager.LockScreenManager import LockScreenManager
 from src.tray import TrayIcon
 from src.backend.Logger import Logger, LoggerConfig, Loglevel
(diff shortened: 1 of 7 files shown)


Core447:
Amazing work!

> reqs.yaml and pypi-requirements.yaml
> 
>     * I'm not sure if these should be updated within each PR, or if that's a packaging job.
> 
>     * I can't seem to generate the pypi-requirements.yaml file. It looks distinct from reqs.yaml in the current upstream commit, and both files only include the command to generate reqs.yaml

You don't need to worry about those. I'm updating them before each release.

> To be resolved in draft: New message format for lock screen settings message
> 
>     * Seeing as there is now support for GNOME, KDE, Cinnamon and Hyprland, I feel like tacking on each supported DE/WM will end up causing a very long text entry if more are added in the future. Looking for feedback for the message format here

Yes, the ideal system for the lock screen detectors (and window grabbers) is not ideal, but it works, and tbh I don't know how we could improve it.

I'm gonna merge the PR, as we can always change stuff later.

PR #383, opened 2025-04-18
Refactor(ActionCore): Removed unneeded GenUI methods

G4PLS:
Some of your methods werent used at all and were overriding the implementations I wrote at some point


G4PLS:
Now when loading the initial UI values we dont update the settings anymore. This only happens at the initial load as we LOAD initial values, so we get them from the settings. There is no need to save them instantly again

Loading the initial ui should only do two things:
1. Load values in the UI
2. Call the callback

gensyn:
I probably won't have time for a review before tuesday.

G4PLS:
> I probably won't have time for a review before tuesday.

Totally understandable, since it's Easter weekend

Core447:
Thank you!

PR #392, opened 2025-05-05
Fixing tray icon not showing up on plasma

### LINKED ISSUE #391: Tray icon fully transparent when running from source on KDE
**Describe the bug**
The tray icon for StreamController is fully transparent when running the application from source (non-Flatpak). It *is* present and can be hovered over, but the icon itself is invisible. This occurs in version `1.5.0-beta.8` (AUR package: `aur/streamcontroller r1620.2b5e328-1`), and has also happened in the past when running the app directly from a cloned Git repository.

**To Reproduce**

1. Install StreamController from AUR (`aur/streamcontroller`) or clone the GitHub repo and run it from source
2. Launch the application on KDE Plasma 6.3.4 (X11 session)
3. Observe the system tray area
4. The tray icon is present but completely transparent

**Expected behavior**
The application should display a visible tray icon in the KDE Plasma system tray while running, just as it does when installed via Flatpak.

**Screenshots**
Not applicable at the moment — I can provide if needed.

**Additional context**

* KDE Plasma version: 6.3.4
* Qt version (system): 6.9.0
* Running under X11
* Installed via AUR (`aur/streamcontroller`)
* Flatpak version does **not** have this issue — tray icon displays correctly
* The icon is interactable and has a hover state, but appears fully transparent
* This also occurred in earlier versions when running directly from the Git repo

Let me know if logs or debug output would be helpful.

ImDevinC:
This resolves #391.

It seems that Plasma is a bit more strict on the GTK requirements for setting application icons.
This change does two things:
1. It puts a copy of the application icon in the Assets/icon folder since we already tell the code
to search that directory.
2. It also adds that same theme path to the Tray Icon so it will pick up icon and display it properly.


Core447:
Thanks!

PR #393, opened 2025-05-06
feat(devcontainers): add devcontainer support to the project

jmylchreest:
Add initial devcontainer support to the project, and some basic linting

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,31 +1,52 @@
 {
-  "name": "StreamController DevContainer",
-  "image": "mcr.microsoft.com/devcontainers/python:3.11",
-  "features": {
-    "ghcr.io/va-h/devcontainers-features/uv:1": {},
-    "ghcr.io/devcontainers-extra/features/apt-packages:1": {
-      "packages": "cmake,libcairo2,libcairo2-dev,pkg-config,libdbus-1-dev,dbus,gir1.2-gtk-3.0,libgirepository1.0-dev"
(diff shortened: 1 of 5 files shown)


mypy.ini
```diff
@@ -0,0 +1,3 @@
+[mypy]
+python_version = 3.11
```
ImDevinC: StreamController is currently using python 3.12

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,31 @@
...
+      "packages": "cmake,libcairo2,libcairo2-dev,pkg-config,libdbus-1-dev,dbus,gir1.2-gtk-3.0,libgirepository1.0-dev"
+    }
+  },
+  "postCreateCommand": "uv venv && . .venv/bin/activate && uv pip install -r requirements.txt && uv pip install black ruff mypy",
+  "customizations": {
+    "vscode": {
+      "settings": {
+        "python.defaultInterpreterPath": "/workspaces/StreamController/.venv/bin/python",
+        "editor.formatOnSave": true,
+        "python.formatter.provider": "black",
```
ImDevinC: This project doesn't use black as the default formatter. Setting this will cause a lot of whitespace changes the first time the save happens. May want to run this by @Core447 to understand if they want a default formatter

Core447: Thanks for catching this! I'm using autopep8, so PEP 8

Core447:
Thank you! Great work!

PR #409, opened 2025-05-28
fix(store): dont fail to load store if a single plugin is misconfigured

ImDevinC:
Currently, if a plugin has a misconfigured manifest or if someone provides an invalid custom repository that is invalid,
the entire store will fail to load and will continue to load indefinitely. This change makes sure that if there is no
valid manifest for a plugin, we just skip it and continue on to the next plugin in the list.


Core447:
Thanks!

PR #410, opened 2025-05-29
[FEAT]: Better Store Badges

G4PLS:
I have reworked the badges to be easy to add and requiring less space.
They also look better but thats just my opinion


G4PLS:
@Core447 would you be able to take a look how to actually scale the images?
I think I tried everything but youre always able to somehow make it scale

src/windows/Store/Icons/IconPage.py
```diff
@@ -83,8 +83,14 @@ def __init__(self, icon_page:IconPage, icon_data:IconData):
         self.set_image(icon_data.image)
         self.set_url(icon_data.github)
 
-        self.set_official(icon_data.official)
-        self.set_verified(icon_data.verified)
+        self.set_official(icon_data.official, """<b>Official Icon Pack</b>
```
ImDevinC: The description should probably be a localized string

src/windows/Store/Plugins/PluginPage.py
```diff
@@ -93,8 +93,19 @@ def __init__(self, plugin_page: PluginPage, plugin_data: PluginData):
         self.set_image(plugin_data.image)
         self.set_url(plugin_data.github)
 
-        self.set_official(plugin_data.official)
-        self.set_verified(plugin_data.verified)
+        self.set_official(plugin_data.official, """<b>Official Plugin</b>
```
ImDevinC: The description should probably be a localized string

src/windows/Store/Preview.py
```diff
@@ -257,21 +254,27 @@ def set_install_state(self, state: int):
...
 
         # Remove ending periods
         if description[-1] == ".":
             description = description[:-1]
-        
-        if len(description) >= 50:
-            description = description[:47] + "..."
-        self.description.set_label(description)
+
+        cutoff = 60
```
ImDevinC, line 260: We went from 50 to 60 here, is that expected?

src/windows/Store/StorePageSection.py
```diff
@@ -78,52 +78,41 @@ def filter_func(self, item):
...
+        name = item.name_label.get_text().lower()
+        author = item.author_label.get_text().lower()
+        description = item.description_label.get_text().lower()
 
-        return fuzz_ratio >= MIN_FUZZY_SCORE
+        name_score = fuzz.ratio(search_string, name)
+        author_score = fuzz.ratio(search_string, author)
+        description_score = fuzz.ratio(search_string, description)
+
+        MIN_FUZZY_SCORE = 20
```
ImDevinC, line 93: Did we intend to change the fuzzy score?

src/windows/Store/Wallpapers/WallpaperPage.py
```diff
@@ -81,8 +81,14 @@ def __init__(self, wallpaper_page:WallpaperPage, wallpaper_data:WallpaperData):
         self.set_image(wallpaper_data.image)
         self.set_url(wallpaper_data.github)
 
-        self.set_official(wallpaper_data.official)
-        self.set_verified(wallpaper_data.verified)
+        self.set_official(wallpaper_data.official, """<b>Official Wallpaper Pack</b>
```
ImDevinC: Should probably use a localized string

src/windows/Store/Preview.py
G4PLS: Yes, this makes the label cutoff a little bit more leanient and doesnt cut off so early. The base window had room to spare horizontally so why not use that

src/windows/Store/StorePageSection.py
G4PLS: Yes, this makes the search a little bit more snappy and makes the biggest difference in the whole search

src/windows/Store/Plugins/PluginPage.py
G4PLS: Localization should now work

src/windows/Store/Wallpapers/WallpaperPage.py
G4PLS: Localization should now work

src/windows/Store/Icons/IconPage.py
G4PLS: Localization should now work

Core447:
I'm not yet completely convinced by the design... I get the idea, but I just think it doesn't really fit the GTK style

Also a problem I'm seeing is that the "verified" state is now a SC icon, which works for now, but if we're adding more users to [this list](https://github.com/StreamController/StreamController-Store/blob/main/OfficialAuthors.json) in the future, it might be weird

Core447:
Reverted the visual changes as the proposed design doesn't fit the GTK style in my opinion.

G4PLS:
@Core447 some things I noticed that you changed back that dont make sense to me:

You removed the whole description change I made on the UI side.
It didnt make a change visually and removed some unneeded boxes

You also changed the cutoff back to what it was but my change there was also working completely fine and just allowing a little bit more space for the short description



Personally I liked the Icons a lot more than the big Text Buttons but as long as the other more important changes stay im fine with it (Meaning mostly the search improvements)

Core447:
@G4PLS 
Didn't meant to revert the description and cutoff change(s). I just did a git revert with some hashes, and I probably did a mistake there or some mistake during merge. Sorry about that.

Can reapply e065242443a63734d0dd94aef744fb505b5d4903 and ecb442920f2337200259c51cdb8de7bec6869711 if those are the (only) ones I accidentally reverted.

G4PLS:
Yeah you only accidentally reverted those two

G4PLS:
But instead of compeltely reverthing the UI changes could we make them smaller?

image
Something in that direction maybe?

My biggest issue is that its just so much wasted space as we only have those 2 badges atm

G4PLS:
Trying for this design to be more in line with the GTK Style
image

G4PLS:
@Core447 @ImDevinC I have now reworked the UI to be more in line with GTK and added proper translations to the Labels and tooltips. If you want you can now review the changes again.
If there is anything to be changed I will take a look at it.

This is now the new UI of the Plugin page, the Icons and Wallpapers look the same but with changed tooltips to match the Tab that were on.
image

G4PLS:
The changes in [150591a](https://github.com/StreamController/StreamController/pull/410/commits/150591a557626189dbbe7c0fbcdd099f02e97f01) was done because I noticed that my Thumbnail was getting cut of by a little bit on the top/bottom.

All thumbnails should follow the 1000x360 resolution so we should make sure that everything from the image is getting shown to the user for a clean user experience

G4PLS:
@Core447 whats the status on this?

Core447:
Thank you!

PR #430, opened 2025-06-17
Adds support for Stream Deck Neo screen

jnfbatista:


bitboxer:
I would really love to have this. What is left to do here?

jnfbatista:
@bitboxer A decision tbh, what I wanted was to just support a clock in that screen, being able to enable disable it, maybe configure alignment and that's all. The decision left to take is:
- Just support a clock and that's it (it could be done quickly enough);
OR
- have it support the existing plugins/follow the plugins approach, so it's more versatile, but I don't even know where to start.

What do you think?

bitboxer:
I would say let's do a) and then later, if there is demand, b). That way there is something working kinda quick-ish. And it might unlock b) later.

jnfbatista:
Sounds good, I’ll let you know when I have something so you can help test it if you’re willing.

Core447:
Thanks for your work and time! But unfortunately, I have to close this in favor of #599

Issue #428, opened 2025-06-16
[BUG] Tray Icon is very large


thecrazytech1:
**Describe the bug**
I am using Arch Linux with hyprland and as my "taskbar" I'm using [ironbar](https://github.com/JakeStanger/ironbar) and when I have StreamController open, the tray icon is quite large

**To Reproduce**
1. Open StreamController with ironbar and its tray icon module enabled

**Expected behavior**
Icon should be smaller

**Screenshots**
![Image](https://github.com/user-attachments/assets/f6580d05-cf31-4230-8bec-64e114c5bf34)

**Additional context**
N/A

ImDevinC:
Should be resolved in beta-1.5.0-beta.12. If not, then we are also tracking some more tray icon improvements in #421

Issue #421, opened 2025-06-06
Tray icon broken on gnome with flatpak


cpw:
**Describe the bug**
The tray icon is broken on Gnome 48 (debian sid) on wayland, when using the flatpak.

Running v1.5.0-beta11

**To Reproduce**
1. Run streamcontroller.
2. Expect tray icon, get three dots instead

**Expected behavior**
A tray icon.

**Screenshots**

**Additional context**
Some logs from journal.

```
Jun 06 09:06:58 cheesypuffs com.core447.StreamController.desktop[29555]: 2025-06-06 09:06:58.220 | DEBUG    | src.backend.PageManagement.Page:load:80 - Loaded page Main in 0.00 seconds
Jun 06 09:06:58 cheesypuffs com.core447.StreamController.desktop[29555]: 2025-06-06 09:06:58.224 | INFO     | src.backend.DeckManagement.DeckController:load_page:666 - Loading page Main on deck CL09J1A01974
Jun 06 09:06:58 cheesypuffs com.core447.StreamController.desktop[29555]: 2025-06-06 09:06:58.226 | INFO     | src.backend.Store.StoreBackend:get_stores:74 - Official store branch: main
Jun 06 09:06:58 cheesypuffs com.core447.StreamController.desktop[29555]: 2025-06-06 09:06:58.305 | INFO     | src.backend.DeckManagement.DeckController:load_background:518 - Loading background in thread: 140675513300672
Jun 06 09:06:58 cheesypuffs com.core447.StreamController.desktop[29555]: 2025-06-06 09:06:58.373 | INFO     | src.backend.DeckManagement.DeckController:load_all_inputs:604 - Loading all inputs took 0.031549692153930664 seconds
Jun 06 09:06:58 cheesypuffs com.core447.StreamController.desktop[29555]: 2025-06-06 09:06:58.389 | INFO     | src.backend.DeckManagement.DeckController:load_page:697 - Loaded page Main on deck CL09J1A01974
Jun 06 09:06:58 cheesypuffs StreamControlle[29555]: Theme parser error: style.css:397:19-27: Unknown name of pseudo-class
Jun 06 09:06:58 cheesypuffs com.core447.StreamController.desktop[29555]: 2025-06-06 09:06:58.428 | DEBUG    | src.backend.DeckManagement.DeckController:update_all_inputs:434 - Updating all inputs took 0.01580023765563965 seconds
[...]

PR #432, opened 2025-06-19
fix(icon): use more distinct path for icon

ImDevinC:
My previous fix for the icon to work on Plasma (#392) broke icon display
on Gnome. It looks as though the appindicator logic in Gnome doesn't
properly scan subdirectories for icons like KDE does (#421). Setting the
full path to our icon allows Gnome appindicator to work, along with 
plasma still working as well. This also fixes the issue with ironbar (#428)
as it also seems to just pick the first icon it can find that matches
the name instead of looking in the correct path.

Tested against the following setups:
- Gnome with appindicator
- hyprland with ironbar
- KDE Plasma
- XFCE X11


Core447:
Thanks!

PR #434, opened 2025-06-19
[FEAT]: Plugin Based Logging

G4PLS:
This adds Plugin Based logging to StreamController.
It adds a log file for every plugin using loguru.

I added custom file retention as Im using date/time based log names that are different on every start. I did this because that makes it easy to only display the latest informations in the about window of the plugin

### DIFF
--- a/GtkHelper/ComboRow.py
+++ b/GtkHelper/ComboRow.py
@@ -27,6 +27,9 @@ class BaseComboRowItem(GObject.GObject):
     def filter_value(self):
         return self.__str__()
 
+    def __repr__(self):
+        return f"<{self.__class__.__name__} display={self.__str__()} value={self.get_value()} filter={self.filter_value}>"
+
 class ComboRowItem(BaseComboRowItem):
     def __init__(self, label: str):
         super().__init__()
--- a/globals.py
+++ b/globals.py
@@ -114,7 +114,6 @@ tray_icon: "TrayIcon" = None
 fallback_font: str = find_fallback_font()
 showed_donate_window: bool = False
 screen_locked: bool = False
-loggers: dict[str, "Logger"] = {}
 
 app_version: str = "1.5.0-beta.11"  # In breaking.feature.fix-state format
 exact_app_version_check: bool = False
--- a/main.py
+++ b/main.py
@@ -100,8 +100,6 @@ def config_logger():
         ]
     )
 
-    Logger.get("PLUGIN").debug("WILLIAM")
-
 class Main:
     def __init__(self, application_id, deck_manager):
         # Launch gtk application
--- a/main.py
+++ b/main.py
@@ -80,26 +80,6 @@ def config_logger():
     log.add(sys.stderr, level="TRACE")
     log.add(write_logs, level="TRACE")
 
-    Logger(
-        LoggerConfig(
-            name="PLUGIN",
-            log_file_path=os.path.join(gl.DATA_PATH, "logs/plugins.log"),
-            base_log_level="TRACE",
-            rotation="3 days",
-            retention=None,
-            compression="zip"
-        ),
-        [
-            Loglevel("TRACE", "trace", 5, "<bold><cyan>"),
-            Loglevel("DEBUG", "debug", 10, "<bold><blue>"),
-            Loglevel("INFO", "info", 20, "<bold><white>"),
-            Loglevel("SUCCESS", "success", 25, "<bold><green>"),
-            Loglevel("WARNING", "warning", 30, "<bold><yellow>"),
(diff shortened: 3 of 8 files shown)


ImDevinC:
I can definitely see benefits here, but I'm also curious if the multiple log files will make it more challenging to grab data from users when they need support. The functionality itself looks good, but will wait for @Core447 to chime in

ImDevinC:
Nevermind, I see my confusion. This doesn't replace the _base_ plugin logger, this creates a new logger specifically for troubleshooting. I think this is a good start, but I think this could be expanded upon a little more to make it more user friendly. Something like the following:

1. On the existing StreamController Plugin Settings page (the one that shows all plugins and gives buttons for settings/about), include a new button for "Troubleshooting". (perhaps we add a new config option to plugins, like `has_troubleshooting` that could determine if this button is enabled)
1. When a user clicks that button, it calls the troubleshoot function of the plugin. This troubleshoot function can do whatever the developer needs (run checks, validate configs, etc) and output the results to a log file (I like the secondary log file for this)
1. After the troubleshooting is completed, we prompt the user to save the log somewhere

I think the only thing missing from your implementation is the ability to call the troubleshoot function from StreamController somewhere. I think this would be better than putting inside a specific action as we may want to troubleshoot some part of the plugin that is preventing actions from loading.

G4PLS:
I want to replace the Plugin Logger with this, its in my opinion a better way and it provides a dedicated "space" for every Plugin to use.
This logger will still write to the main log and the CLI but has a separate log file for the Plugin to easily get everything thats going on with the Plugin without having to read through sometimes bigger log files with lots of unneeded entries.

Adding the button to the settings that starts a troubleshoot/diagnostic sounds like a good idea!

Under `Settings -> Plugins -> Your Plugin -> About -> Troubleshooting -> Debugging Info` we already have the section to show the log, just like StreamController does it and I would just use that.

I would create 2 methods:
`log()` which will replace the new `troubleshoot()` method. This can then be used to just log things into that Plugin log.
I will create a new `troubleshoot()` method that can be called to implement custom diagnostic behaviour, I would do that In a way where I then call a diagnostic on all Actions to get as much data as needed by using signals.

G4PLS:
@ImDevinC could you take a look again? Added everything you wanted/needed

G4PLS:
This change now results in the following (Its a long line):

```
2025-06-20 15:19:52.578 | com_gapls_AudioControl | actions.AudioCore:troubleshoot:190 - Troubleshooting for action: Mute on  : {'selected_device': <Device pulse_name=Browser pulse_index=34 device_name=Browser>, 'device_filter': <SimpleComboRowItem display=Sink value=sink filter=Sink>, 'info_content': <SimpleComboRowItem display=Volume value=volume filter=Volume>, 'show_info_content': True, 'device_name': '', 'show_device_name': True, 'use_standard': False, 'loaded_devices': [<Device pulse_name=System pulse_index=31 device_name=System>, <Device pulse_name=Game pulse_index=32 device_name=Game>, <Device pulse_name=Voice Chat pulse_index=33 device_name=Voice Chat>, <Device pulse_name=Browser pulse_index=34 device_name=Browser>, <Device pulse_name=alsa_output.pci-0000_18_00.6.analog-stereo pulse_index=75 device_name=ALC897 Analog>, <Device pulse_name=alsa_output.usb-Focusrite_Scarlett_Solo_USB_Y756R404B041C6-00.HiFi__Line1__sink pulse_index=79 device_name=Scarlett Solo USB>], 'icon_keys': [<Icons.MUTED: 'mute'>, <Icons.UNMUTED: 'audio'>], 'current_icon_name': <Icons.UNMUTED: 'audio'>}
```

The code for this is very simple:
```py
async def troubleshoot(self, *args, **kwargs):
    self.log(f"Troubleshooting for action: {self.__class__.__name__} on ",
             selected_device=self.selected_device,
             device_filter=self.device_filter,
             info_content=self.info_content,
             show_info_content=self.show_info_content,
             device_name=self.device_nick,
             show_device_name=self.show_device_name,
             use_standard=self.use_standard_device,
             loaded_devices=self.loaded_devices,
             icon_keys=self.icon_keys,
             current_icon_name=self._icon_name)
```

ImDevinC:
This looks great, thanks for putting it together so quickly!

G4PLS:
Alright opened it for merge! Now we just have to wait and see what @Core447 thinks

ImDevinC (approved):

G4PLS:
@Core447 bumping this aswell.

what do you think?

Core447:
I like it. Only reason I'm not merging yet is a very selfish one: I don't have very much experience with all things logging, and I want to try a bit around with it, so that I understand it better and can work with this myself. 

Hope I can merge soon

PR #443, opened 2025-06-27
Scroll delay

gensyn:
I added a delay at the beginning and end of scrolling to make it look smoother.

output


Core447:
Thank you!

PR #452, opened 2025-07-10
chore(dependencies): Updating flatpak dependencies to match requirements.txt

ImDevinC:
When we updated dependencies [here](https://github.com/StreamController/StreamController/commit/e92b7f7711bb20732ea0b0a992d8d2ee37f31723), forgot to update flatpak dependencies.

I also deleted `reqs.yaml` as the flatpak builder references `pypi-requirements.yaml` and it was a bit confusing. I did a quick check, but I don't see that file being used at all anywhere.


Core447:
Thank you, but I'm gonna close this, as I'm usually updating `requirements.txt` and also the `pypi-requirements.yml` before each release.

PR #455, opened 2025-07-19
feat: add option to set brightness to zero on lock screen

JoeyEamigh:
This is a very simple PR that adds an option to set the brightness to zero on the lock screen rather than turning on the screensaver.


Core447:
Is there any particular reason for avoiding the screensaver to achieve this?

Core447:
Closing this, because the brightness can already be set to 0 in the screensaver settings

PR #468, opened 2025-08-05
Auto-update plugins only if commit timestamp is after local timestamp

gensyn:
This has been bothering me for a while. If I make local changes to my plugin and forget to disable auto-update for plugins, StreamController resets my repository to the commit from the store. With these changes, plugins are only updated if the commit from the store was done after the local commit. This probably only benefits developers but still makes sense.

### DIFF
--- a/src/backend/Store/StoreBackend.py
+++ b/src/backend/Store/StoreBackend.py
@@ -57,6 +57,7 @@ class StoreBackend:
     STORE_CACHE_PATH = "Store/cache"
     # STORE_CACHE_PATH = os.path.join(gl.DATA_PATH, STORE_CACHE_PATH)
     STORE_BRANCH = "1.5.0"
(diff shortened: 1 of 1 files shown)


ImDevinC:
I'm not sure i agree with this. If i have auto update enabled, i would expect it to match the store version always

Core447:
I see the point of both of you, but as there's already the `--devel` flag for developers which also disables auto store updates, I would close this PR if you're ok with it @gensyn ?

Edit: Also one problem I see with this PR is what happens if someone works on a plugin, but the original author updates the one in the store and therefore the remote timestamp. This will then result in the local changes of the other person being overwritten.

gensyn:
I didn't know about the `--devel` flag. Doesn't really solve my problem though, as I would like to auto update everything that I am not working on. But it's been so long now... just close it.

Core447:
> But it's been so long now... just close it.

Alright. I never had this problem, cause I have a normal installation and a separate dev version (via the --data flag)

PR #482, opened 2025-10-12
Load icon from non-scalable path

razzeee:
The scalable paths should expect a svg icon, but we only have a .png so we need to load it from a non-scalable path

I had to transform the icon slightly as the one from the repo wasn't a square. 

Do you have the source files for this image saved somewhere or should we backup the higher resolution image from before somewhere?

Might help with #421


ImDevinC:
https://specifications.freedesktop.org/icon-theme-spec/latest/#install_icons
This mentions we should at minimum have a 48x48 icon in `hicolor/apps`. I was in the process of creating a similar PR, but since this is here already, can you also add that this PR? I also think if we do this, we may be able to remove the [set_icon](https://github.com/StreamController/StreamController/pull/482/files#diff-9a6a201330233e0ed514f904f8642d53bd4697c11178047850d3fbc7ddf1aaefL21) line entirely as it should try to load the icon organically

As for keeping the image, the larger image is already available in flatpak/icon.png, so I don't know if we need another backup unless I'm misunderstanding

razzeee:
@ImDevinC done, but untested

also a bit confused about the wording you've linked too - will investigate that. but it shouldn't hurt to have the 48x48 so, whatever :)

ImDevinC (approved):

ImDevinC:
Thanks for the quick fix! I can rebuild the flatpak this week and test this change

ImDevinC:
The specific line I'm quoting from the link is this part

> Minimally you should install a 48x48 icon in the hicolor theme. This means installing a PNG file in $prefix/share/icons/hicolor/48x48/app

Having the 512x512 is fine too, but we should _at least_ have this. Admittedly I also missed this when I made the first PR to fix tray icons as I shuold've caught it then. Thanks for addressing it

razzeee:
No, this is completely untested and done blindly :)

razzeee:
> The specific line I'm quoting from the link is this part
> 
> > Minimally you should install a 48x48 icon in the hicolor theme. This means installing a PNG file in $prefix/share/icons/hicolor/48x48/app
> 
> Having the 512x512 is fine too, but we should _at least_ have this. Admittedly I also missed this when I made the first PR to fix tray icons as I shuold've caught it then. Thanks for addressing it

I do think minimally could mean two things here:

- We need to always have a 48x48
- 48x48 is the smallest available, bigger is better

In most apps I just have .svg and it's fine - so I don't really know how non scalable works

razzeee:
Btw the minimum on flathub right now is 128x128 - but we prefer to have bigger icon images

razzeee:
IMO, we should be fine without the 48x48

ImDevinC:
Based on the spec, we should be including 48x48, no matter what. But we can include the other versions as well as they give more options, doesn't hurt.

ImDevinC:
Built the flatpak and this doesn't work. I get a green alien icon. Doing some testing to see what change broke it

src/tray.py
```diff
@@ -17,8 +20,9 @@ def __init__(self):
         self.menu.add_menu_item(5, "About", callback=self.on_about)
         self.menu.add_menu_item(6, menu_type="separator")
         self.menu.add_menu_item(7, "Quit", callback=self.on_quit)
-        super().__init__(self.menu, self.MenuPath, self.IndicatorPath, self.AppId, "StreamController")
-        self.set_icon("com.core447.StreamController", os.path.join(gl.top_level_dir, "Assets", "icons", "hicolor", "scalable", "apps"))
+        super().__init__(
+            self.menu, self.MenuPath, self.IndicatorPath, self.AppId, "StreamController"
+        )
```
ImDevinC:
```suggestion
        )
        self.set_icon("com.core447.StreamController")
```

ImDevinC: We can use the app name to set the icon now that we have proper icons in place. Without this change, the icon just loads as a green alien icon (or some other default)

ImDevinC: Confirmed that with this change the flatpak loads the icon properly in both Gnome with Appindicator extension installed, and on Hyprland with a tray icon extension installed.

razzeee:
Do you want this change as an additional commit or should I rebase and squash everything?
(I would also remove those unneeded formatting changes then)

ImDevinC:
Squashing is probably best.

razzeee:
Should be squashed and cleaned up now

razzeee:
Is this waiting for something specific?

Core447:
@ImDevinC Just to confirm: Did this work in your tests?

ImDevinC:
@Core447 yep, it worked as described

PR #490, opened 2025-10-29
fix(store): Check status code when receiving a response

ImDevinC:
The requests library does not throw an exception on all non 2xx codes, for example 429. This seems to be happening frequently from GitHub so
this change is designed to make sure we validate the status code, and if it is not what we expect, return an error that the rest of the 
application already knows how to handle


Core447:
Thank you!

PR #491, opened 2025-10-29
fix(store): If attribution.json is missing, don't prevent loading of plugin

ImDevinC:
The previous fix in #490 caused caused an issue where if attribution.json is missing,
the plugin would fail to load in the store. Since attribution.json is not required,
this makes sure to return an empty value, which is valid


Core447:
Thanks for the quick fix!

PR #495, opened 2025-11-16
Fix shutdown issues

ImDevinC:
I've noticed that when the app is shutting down, it often hangs for a few seconds, sometimes upwards of 10. This change helps address that by doing the following:
- Call the safe `stop_monitoring()` function of `USBMonitor`
- Call the safe `stop()` function of `loguru`
- Ignore some specific threads that we don't want/can't terminate

Doing some initial testing, StreamController now quits within 2 seconds during most of my testing

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -184,14 +184,15 @@ class App(Adw.Application):
         for ctrl in gl.deck_manager.deck_controller:
             ctrl.delete()
 
+        gl.deck_manager.stop_usb_monitoring()
+
         gl.plugin_manager.loop_daemon = False
-        log.debug("non-daemon threads:")
+
         for thread in threading.enumerate():
-            if thread is not threading.current_thread():
-                thread.join(timeout=2.0)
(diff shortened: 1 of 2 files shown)


src/app.py
```diff
@@ -184,14 +184,18 @@ def on_quit(self, *args):
...
         gl.plugin_manager.loop_daemon = False
-        log.debug("non-daemon threads:")
+
         for thread in threading.enumerate():
-            if thread is not threading.current_thread():
-                thread.join(timeout=2.0)
+            if (thread is not threading.current_thread()
+                and "loguru-" not in thread.name  # Loguru thread is manually cleaned up later
+                and "force_quit_timer" != thread.name  # This is needed to make sure we force quit
+                and "Thread-1" != thread.name):  # Thread-1 is a Gtk application thread that can't quit
```
ImDevinC: Ugh... immediately after submitting this PR, this changed so I think there's something else at play here that i need to investigate still

Core447:
This is amazing! Thank you!

PR #496, opened 2025-11-17
Fix Stream Deck Plus Swipe Event Detection

designgears:
# Fix Stream Deck Plus Swipe Event Detection

## Problem
Swipe gestures on the Stream Deck Plus touchscreen were not being detected.

## Root Causes
1. **Inverted Direction Logic**: Left swipe triggered right event and vice versa  
2. **Missing Dial Support**: Swipe events weren't sent to dial inputs (unlike touch press events)

## Solution
- Corrected swipe direction logic in [ControllerTouchScreen.event_callback()](cci:1://file:///tmp/StreamController/src/backend/DeckManagement/DeckController.py:2340:4-2385:25)
- Added swipe event forwarding to corresponding dial inputs 

## Result
✅ Swipe gestures now work correctly on Stream Deck Plus touchscreen
✅ Events are routed to both touchscreen and dial inputs  
✅ Actions can be configured to respond to `DRAG_LEFT`/`DRAG_RIGHT` events

### DIFF
--- a/src/backend/PageManagement/PageManagerBackend.py
+++ b/src/backend/PageManagement/PageManagerBackend.py
@@ -323,7 +323,7 @@ class PageManagerBackend:
 
         return list(pages_set)
 
-    def reload_pages_with_path(self, path: str):
+    def reload_pages_with_path(self, path: str, brightness: bool = True, screensaver: bool = True, background: bool = True, inputs: bool = True):
         pages = self.get_pages_with_path(path)
 
         for page in pages:
@@ -332,7 +332,11 @@ class PageManagerBackend:
             if page.deck_controller.active_page != page:
                 continue
 
-            page.deck_controller.load_page(page, allow_relaod=True)
+            page.deck_controller.load_page(page, allow_relaod=True,
+                                           load_brightness=brightness,
+                                           load_screensaver=screensaver,
+                                           load_background=background,
+                                           load_inputs=inputs)
 
     @staticmethod
     def reload_all_pages() -> None:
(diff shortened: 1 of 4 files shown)


Core447:
> 2\. **Missing Dial Support**: Swipe events weren't sent to dial inputs (unlike touch press events)

This is actually intended. The swipe events can be used by adding actions to the screen itself. The idea behind it is, that in most cases the swipes will probably be used to switch pages (I believe this is also how it works in the official Elgato app). And it therefore doesn't really make sense to send the events to the dials.

But if you see a use-case for receiving the events on the dials, let me know.


Edit: Just noticed that you mention the events now get send to both the dials and the touchscreen. This way users could still have page switch actions on the touchscreen itself. But for what do you need the swipe events on the dials?

designgears:
> > 2. **Missing Dial Support**: Swipe events weren't sent to dial inputs (unlike touch press events)
> But if you see a use-case for receiving the events on the dials, let me know.
> 
> Edit: Just noticed that you mention the events now get send to both the dials and the touchscreen. This way users could still have page switch actions on the touchscreen itself. But for what do you need the swipe events on the dials?

I'm building a plugin for pipeweaver, I planned on adding some left/right swiping so I could add some per dial screen paging instead of full screen paging. 

I think both can co-exist, the current swipes can be executed on a single dial screen, there could be a longer swipe for full page changes.

designgears:
I just realized this could be odd if you don't have a stream deck plus, while you can use the entire screen and that's how stream controller handles it, each dial is also a quarter of the screen. This PR gives you the ability to swipe on the quadrants rather than the whole screen.

Edit: the idea of the co-existing, require the swipe for changing pages to be at least X number of pixels or make contact with at least 3 of the of the quadrants to trigger it. That would allow small swipes on each quadrant and long swipes for page changes.

Core447:
I don't have a SD+ for testing: Does the StreamDeck registers the swipe if it is only one 1/4 of the screen?

Didn't look at the code in detail yet, just to confirm: you made it so that every swipe gets send to the actions on the Screen, but also to the actions of the dial that matches the swipe position?

mike3nl:
> I don't have a SD+ for testing: Does the StreamDeck registers the swipe if it is only one 1/4 of the screen?
> 
> Didn't look at the code in detail yet, just to confirm: you made it so that every swipe gets send to the actions on the Screen, but also to the actions of the dial that matches the swipe position?

I could help. I habe a new sd+ 3 weeks old. And i have done a lot with it.
But i'm not a python guy. My field of work is the Industrial Automation like Airbus, ESA, GPL and Nasa and all in C++.

Core447:
> My field of work is the Industrial Automation like Airbus, ESA, GPL and Nasa and all in C++.

Sounds very interesting!

And thanks for the offer of testing, but let's wait a couple of days for @designgears as he already has the dev environment set up.

jeinstei:
Also interested and happy to test as needed.

designgears:
I canned my features that were going to use this, but I can whip up a small plugin to demonstrate and let others test it out as well.

PR #502, opened 2025-12-05
fix(ui): Prevent background color being incorrectly written during page load

Pol-Valentin:
## Summary
- Fix background color `[0,0,0,0]` being incorrectly written to page config during page navigation
- Replace unreliable `disconnect_by_func`/`connect_signals` pattern with GTK-recommended `handler_block`/`handler_unblock`
- Fix potential signal leak when `load_for_identifier` returns early

## Problem
When navigating between pages (e.g., using a deck button to go back/forward), the background color for the key at position (0,0) was sometimes being overwritten with `[0,0,0,0]` (transparent), causing the background to disappear.

## Root Cause
`ColorDialogButton`'s `notify::rgba` signal fires for **all** changes including programmatic ones via `set_rgba()`, unlike `ColorButton`'s `color-set` signal which only fires on user interaction.

The existing `disconnect_by_func()` approach was unreliable because:
1. It can fail silently with Python bound methods ([PyGObject issue #106](https://gitlab.gnome.org/GNOME/pygobject/-/issues/106))
2. GTK signals may be queued and processed even after disconnection

## Solution
Use `handler_block(handler_id)` / `handler_unblock(handler_id)` with stored handler IDs, which is the [GTK-recommended approach](https://docs.gtk.org/gobject/func.signal_handler_block.html) for temporarily suppressing signal handlers.

## Test plan
- [ ] Remove any existing `"color": [0,0,0,0]` from a page config file for key 0x0
- [ ] Launch the app and navigate between pages multiple times using deck buttons
- [ ] Verify the color is not incorrectly written back to the config
- [ ] Verify manually changing background color via the UI still works
- [ ] Verify the revert button still works

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/windows/mainWindow/elements/Sidebar/elements/BackgroundEditor.py
+++ b/src/windows/mainWindow/elements/Sidebar/elements/BackgroundEditor.py
@@ -109,14 +109,16 @@ class ColorRow(Adw.PreferencesRow):
         self.connect_signals()
 
     def connect_signals(self):
-        self.button.button.connect("notify::rgba", self.on_change_color)
-        self.button.revert_button.connect("clicked", self.on_revert)
+        self._color_handler_id = self.button.button.connect("notify::rgba", self.on_change_color)
+        self._revert_handler_id = self.button.revert_button.connect("clicked", self.on_revert)
 
-    def disconnect_signals(self):
-        try:
-            self.button.button.disconnect_by_func(self.on_change_color)
-        except:
-            pass
+    def block_signals(self):
+        if hasattr(self, '_color_handler_id') and self._color_handler_id:
+            self.button.button.handler_block(self._color_handler_id)
+
+    def unblock_signals(self):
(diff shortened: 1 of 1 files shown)


ImDevinC:
@Pol-Valentin do the block/unblock changes directly impact the background color changes you're addressing here? It seems unrelated at first glance, and not something we need to change.

Core447:
Didn't know about the block and unblock methods. But as far as I understand the docs, disconnect_by_func should work as well.

https://gitlab.gnome.org/GNOME/pygobject/-/issues/106 seems to only be relevant if we would want to remove multiple handlers at once. But normally we should only have one.

Not saying your fix doesn't work, but interest in why it works. Only idea I have is that maybe it can happen that we mistakenly create a new handler, when there's still one connected , which then stay connected forever.

Core447:
Closing this for now, as I don't think this really solves the problem.

PR #508, opened 2025-12-28
Adding mk.2 deck to udev rules

ImDevinC:
This deck is already supported in the code, but the udev rules
never got updated so it causes confusion


hamid-elaosta:
I have a MK.2 which reports a different PID, might be worth adding that too?

```
Bus 005 Device 002: ID 0fd9:0080 Elgato Systems GmbH Stream Deck MK.2
```

The box says:
SKU: 10GBA9901
RMN: 20GBA9901
Part: 41GBA9909

ImDevinC:
@hamid-elaosta that device is already listed: https://github.com/StreamController/StreamController/blob/main/udev.rules#L7

hamid-elaosta:
My bad, seems I switched my brain off for the Christmas holidays.

Core447:
Thank you!

PR #509, opened 2025-12-29
fix(KeyGrid): Set fixed size and pixel size for key button images

designgears:
Before:
image

After
image

### DIFF
--- a/src/windows/mainWindow/elements/KeyGrid.py
+++ b/src/windows/mainWindow/elements/KeyGrid.py
@@ -142,6 +142,8 @@ class KeyButton(Gtk.Frame):
 
         self.image = Gtk.Image(hexpand=True, vexpand=True, css_classes=["key-image", "key-button"])
         self.image.set_overflow(Gtk.Overflow.HIDDEN)
+        self.image.set_size_request(75, 75)
+        self.image.set_pixel_size(75)
         self.set_child(self.image)
 
         # self.button.connect("clicked", self.on_click)


ImDevinC:
Couple things:
1. This should be a `fix` not a `chore`
2. `72` may not always be the right pixel size. Luckily, we have the expected pixel size for each deck already stored on the deck object as seen here: https://github.com/StreamController/streamcontroller-python-elgato-streamdeck/blob/master/src/StreamDeck/Devices/StreamDeck.py#L440. We should use this value instead of hardcoding a value

designgears:
> Couple things:
> 
> 1. This should be a `fix` not a `chore`
> 2. `72` may not always be the right pixel size. Luckily, we have the expected pixel size for each deck already stored on the deck object as seen here: https://github.com/StreamController/streamcontroller-python-elgato-streamdeck/blob/master/src/StreamDeck/Devices/StreamDeck.py#L440. We should use this value instead of hardcoding a value

1. got me there, I messed that up twice. 
2. From my understanding this only changes the display size in the app, if we use the values from `https://github.com/StreamController/streamcontroller-python-elgato-streamdeck/blob/master/src/StreamDeck/Devices/StreamDeck.py#L440` things end up a bit wonky. 

image

ImDevinC:
Thanks for clarifying that second point!

Core447:
Thanks to you both!

designgears:
> 2. `72` may not always be the right pixel size. Luckily, we have the expected pixel size for each deck already stored on the deck object as seen here: https://github.com/StreamController/streamcontroller-python-elgato-streamdeck/blob/master/src/StreamDeck/Devices/StreamDeck.py#L440. We should use this value instead of hardcoding a value

Was thinking about this a bit more, we should use the real sizes for everything and just scale the software preview, that would be more accurate to see issues, like hard coding tiles sizes in an plugin would be obvious.

Core447:
> Was thinking about this a bit more, we should use the real sizes for everything and just scale the software preview, that would be more accurate to see issues, like hard coding tiles sizes in an plugin would be obvious.

Didn't look at it in detail, but isn't this already the case? Cause your PR basically only sets the pixel size for the UI? We still give GTK the images in key-resolution, don't we?

designgears:
> > Was thinking about this a bit more, we should use the real sizes for everything and just scale the software preview, that would be more accurate to see issues, like hard coding tiles sizes in an plugin would be obvious.
> 
> Didn't look at it in detail, but isn't this already the case? Cause your PR basically only sets the pixel size for the UI? We still give GTK the images in key-resolution, don't we?

That's the case on the hardware side, I'm suggesting we do the same in software and scale the UI preview. That way it would be accurate for testing a plugin across virtual devices for basic things like geometry and such.  

Edit: I see the fake deck is hard coded to the plus, so that would need some work too.

designgears:
Created a PR to do all of this! 
https://github.com/StreamController/StreamController/pull/512

PR #512, opened 2026-01-02
Enhance touchscreen handling, fake deck configuration, and UI styling

designgears:
## Summary

This PR improves touchscreen handling, enhances fake deck configuration capabilities, fixes UI styling to accurately match physical hardware, and resolves a GObject Introspection warning.

### 1. Enhance touchscreen handling and fake deck configuration
- **Added safety checks** for touchscreen size to prevent errors when touchscreen is not available
- **Updated DeckController and related classes** to return `None` for touchscreen size when not applicable
- **Improved fake deck loading logic** to allow for customizable deck types based on settings
- **Refactored settings page** to dynamically update deck type selections based on the number of fake decks configured

**Files changed:**
- `src/backend/DeckManagement/DeckController.py`
- `src/backend/DeckManagement/DeckManager.py`
- `src/backend/DeckManagement/Subclasses/FakeDeck.py`
- `src/windows/Settings/Settings.py`
- `src/windows/mainWindow/DeckPlus/ScreenBar.py`
- `src/windows/mainWindow/elements/KeyGrid.py`

### 2. Fix GObject Introspection warning
- Added `gi.require_version("Gdk", "4.0")` to `HelperMethods.py` to resolve GObject Introspection version warning

**Files changed:**
- `src/backend/DeckManagement/HelperMethods.py`

## Testing

- [x] Verified touchscreen handling works correctly when touchscreen is available
- [x] Verified no errors occur when touchscreen is not available
- [x] Tested fake deck configuration with different deck types
- [x] Verified settings page correctly updates deck type selections
- [x] Verified no GObject Introspection warnings appear

## Impact

- **Improved reliability**: Prevents errors when touchscreen is unavailable
- **Enhanced flexibility**: Fake deck configuration is more customizable
- **Code quality**: Resolved GObject Introspection warning

## Images

Before: 
image

After:
image

Fake Decks: 
image
image
image
image


Core447:
First of all, thank you!
Some things I noticed:
- For the change in Fake Deck Type a restart of the app is required which is a bit annoying. Especially because it isn't noted anywhere. I know that the app doesn't handle it well if we change the deck configuration on the fly. But maybe we can internally remove the old deck and then add the new one on each change?
- For consistency reasons the keys in the UI should be the same size across the different devices. (I know that they differ on the physical devices, but still).
- I understand that you prefer black as the background for the keys in the UI, as this is again how it is on the physical deck. However, the old grey looks way cleaner in the app. And I don't see a case where the user would get something from having the black backgrounds. But I actually see a use case where the transparent background is neat: To see if the action uses a transparent background or not. This allows the user to quickly see if a wallpaper or background color would be visible on a key.
- You replaced all "&" with "and" in locales.csv, even for other languages than English.

I really like that it's not possible to configure the type of fake decks! Great work!

designgears:
> First of all, thank you! Some things I noticed:
> 
> * For the change in Fake Deck Type a restart of the app is required which is a bit annoying. Especially because it isn't noted anywhere. I know that the app doesn't handle it well if we change the deck configuration on the fly. But maybe we can internally remove the old deck and then add the new one on each change?
> * For consistency reasons the keys in the UI should be the same size across the different devices. (I know that they differ on the physical devices, but still).
> * I understand that you prefer black as the background for the keys in the UI, as this is again how it is on the physical deck. However, the old grey looks way cleaner in the app. And I don't see a case where the user would get something from having the black backgrounds. But I actually see a use case where the transparent background is neat: To see if the action uses a transparent background or not. This allows the user to quickly see if a wallpaper or background color would be visible on a key.
> * You replaced all "&" with "and" in locales.csv, even for other languages than English.
> 
> I really like that it's not possible to configure the type of fake decks! Great work!

1. I noticed this as well and was trying to work around it, it's pretty odd, you can add them all, change the type, remove them all, and add them again and it works. Even before this PR things can end up very slow and a restart is needed to correct that. 
2. Selfish reasons on my part for this one since the keys are larger on the SD+, it's very nice seeing the full resolution of them and the touchbar. An option to render them to the full size would be a good compromise I think. 
3. Selfish reasons again here, it made finding rendering issues easier, but then again I swapped thru a lot of colors while testing. This we can omit, setting the background color per button or deck wide is sufficient. 
4. Didn't think that thru, oof. 

This was all rushed together just as a proof of concept, I'll keep hammering on the fake deck type swap and see if I can make that play nice. 

Edit: reworked what I was trying to do before, figured out my mistake and made that work pretty well!

PR #515, opened 2026-01-03
Fix Localization Issues: HTML Entity Escaping & CSV Parsing

designgears:
This PR fixes two critical localization issues that were causing display problems in the UI.

### Issues Fixed

1. **GTK Markup Parsing Errors**: Unescaped ampersands and other HTML entities in translated strings were causing GTK warnings and rendering failures
2. **Language Display Mismatch**: German text with semicolons was being incorrectly parsed, causing English UI to show German text

This ensures robust localization handling for all languages and prevents future markup-related display issues.

Before:
image

image

After:
image
image

### DIFF
--- a/locales/LocaleManager.py
+++ b/locales/LocaleManager.py
@@ -16,6 +16,7 @@ along with this program. If not, see <https://www.gnu.org/licenses/>.
 import os
 import csv
 import locale
+import html
 from loguru import logger as log
 
 class LocaleManager:
@@ -89,5 +90,9 @@ class LocaleManager:
 
         if result is None:
             result = fallback
+
+        # Escape HTML/XML entities for GTK markup compatibility
+        if result:
+            result = html.escape(result)
 
         return result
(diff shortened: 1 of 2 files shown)


locales/LocaleManager.py
```diff
@@ -90,4 +91,8 @@ def get(self, key: str, fallback: str = None) -> str:
         if result is None:
             result = fallback
 
+        # Escape HTML/XML entities for GTK markup compatibility
```
ImDevinC, line 94: Is there a specific place this fixed something? I don't see it in the screenshots but I'm on mobile so maybe I'm missing it. Just concerned that this might need testing through the entire app, as we do actually use HTML in some places to link out

designgears: Sorry, guess I should have explained the screenshots a bit better. The ampersands break the title on the performance page and anywhere else they get used the text label will simply be missing unless escaped.

Core447:
Nice!

PR #518, opened 2026-01-03
Window Grabber Supported Sway Environments

jgoett154:
I noticed recently that Sway/SwayFX returns different values for XDG_CURRENT_DESKTOP now vs when the integration was first added. This just adds in 2 supported envs for Sway so that it covers Sway itself as well as the popular SwayFX fork. Left the original sway option just in case it is still used or set manually.


Core447:
Thank you!

PR #526, opened 2026-01-07
[Bug Fix]: Stabilize resume by keeping controllers alive during transient HID errors

getjump:
### Background
After suspend/resume, Stream Deck HID handles can temporarily fail (“No HID device”), even with the `Use new resume mode (beta)` flag turned ON. The previous
behavior removed controllers after repeated touchscreen transport errors, which triggered re-enumeration against
devices that were not yet ready and led to crashes/races.

### Changes:

1. Keep controllers alive on repeated touchscreen transport errors instead of removing/reconnecting.
2. Guard update_all_inputs() and tick_actions() when active_page is None to avoid resume-time crashes.

### Result
The app now survives transient HID failures and recovers without process restarts or race-condition
exceptions.

### DIFF
--- a/src/backend/DeckManagement/DeckController.py
+++ b/src/backend/DeckManagement/DeckController.py
@@ -98,14 +98,11 @@ class MediaPlayerSetTouchscreenImageTask:
             log.error(f"Failed to set deck touchscreen image. Error: {e}")
             MediaPlayerSetTouchscreenImageTask.n_failed_in_row += 1
             if MediaPlayerSetTouchscreenImageTask.n_failed_in_row > 5:
-                log.debug(f"Failed to set touchscreen image for 5 times in a row for deck {self.deck_controller.serial_number()}. Removing controller")
-                
-                
-                self.deck_controller.deck.close()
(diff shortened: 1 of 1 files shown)


src/backend/DeckManagement/DeckController.py
```diff
@@ -98,14 +98,11 @@ def run(self):
             log.error(f"Failed to set deck touchscreen image. Error: {e}")
             MediaPlayerSetTouchscreenImageTask.n_failed_in_row += 1
             if MediaPlayerSetTouchscreenImageTask.n_failed_in_row > 5:
-                log.debug(f"Failed to set touchscreen image for 5 times in a row for deck {self.deck_controller.serial_number()}. Removing controller")
```
ImDevinC, line 101: I don't know if we want to remove the ability to entirely reset the decks. If the error is real, and not transient, we would never attempt to reset the deck. One could make an argument for bumping this up from 5 to higher number possibly, but as I've never experience this issue will need to rely on @Core447 for input

Core447: @getjump Also never had problems with the resume mode. But your fix should still work, if we keep the original logic where it removes the deck after too many errors, right? But as ImDevinC suggested, we could bump the number to maybe 10?

PR #531, opened 2026-01-11
update Levenshtein

### LINKED ISSUE #529: Arch update to python 3.14 breaks streamcontroller
Just a warning to arch users to avoid updating python for now - or live with streamcontroller breaking for a bit.

python: error while loading shared libraries: libpython3.13.so.1.0: cannot open shared object file: No such file or directory

ImDevinC:
Fixes #529


reqs.yaml
ImDevinC, line 1: This is just a dup of pypi-requirements.yaml, removing the duplicated file

ImDevinC:
Looks like #532 already covers this and has been fixed , closing this

PR #532, opened 2026-01-11
Bump RapidFuzz==3.14.3 and typing_extensions==4.14.1 to support Python 3.14

getjump:


getjump:
For some reason
```python -m req2flatpak --requirements-file requirements.txt --yaml -o reqs.yaml --target-platforms 312-aarch64 312-x86_64```
generates empty `reqs.yaml` for me

getjump:
It seems like a regression inside req2flatpak 

- python -m req2flatpak ... -o /tmp/test.yaml → produces empty file
- python -m req2flatpak ... > /tmp/test.yaml → produces correct YAML

This was fixed in upstream https://github.com/johannesjh/req2flatpak/commit/c25928b229e35d5e0202c69ce72b6f9d704eae8f but unfortunately there are no release with this fix yet

I have regenerated this files and have updated the header to actual command

nosduco:
@Core447 If possible, getting this in would fix https://aur.archlinux.org/packages/streamcontroller without having to add patches to update requirements in PKGBUILD. If this will take a while I can add a patch though.

Core447:
> @Core447 If possible, getting this in would fix https://aur.archlinux.org/packages/streamcontroller without having to add patches to update requirements in PKGBUILD. If this will take a while I can add a patch though. 

Sure. Hope its going to work now

getjump:
@nosduco `streamcontroller-git` from AUR is working for me now, so you can upgrade `streamcontroller` without a problem

PR #537, opened 2026-01-19
fix: kdotools search needs a search term

jlskuz:
Calling just "kdotool search" fails with "Error: missing argument". According to the documentation it expects a search term. The term [can be a regular expression](https://github.com/jinliu/kdotool/blob/master/src/templates.rs#L92C11-L92C13), hence we use "." to search for any character as we want to list all windows.


Core447:
Thank you!

PR #539, opened 2026-01-21
Adding support for dials that don't have an associated touchscreen

jeinstei:
Adding support for dials in a DialBox that don't have a touch screen to avoid error messages an breaking plugins when testing new devices. Separate PR will be created to handle custom layouts, hopefully in a sustainable manner.


jeinstei:
Sorry -- was just double-checking the SD+ code and how StreamController calls TOUCH

PR #546, opened 2026-02-01
[draft] Add support for n3 style devices

jeinstei:
Merging support for N3 layout and a framework for possibly adding other layouts.

image


jeinstei:
I would like some thoughts from the maintainers as to how you want to handle such things, if at all. This could be a plugin model, with a hook in DeckConfig, a check like I am doing, or a few different things depending on how layout is also included in.

jeinstei:
An option would be to have a list of "layout" classes that are matched, instead of the current DeckConfig architecture.

jeinstei:
So lets try changing this to a "layout" class, such that plugins could even register their own layout. Some sort of a dictionary with "default" being the usual layout, and others can be added and such.

jeinstei:
- [ ] Convert layout system to allow for class selection based on device (somehow) for non SD layouts
- [ ] Determine "Standard" layout components (dials, dials w/ touchscreens, keys, etc) and better abstract
- [ ] Determine how this interacts with lower-level control library
- [ ] Are new devices able to be plugins? Add "layout registration" feature

main.py
```diff
@@ -210,22 +211,22 @@ def update_assets():
 @log.catch
 def reset_all_decks():
     # Find all USB devices
-    devices = usb.core.find(find_all=True, idVendor=DeviceManager.USB_VID_ELGATO)
+    devices = usb.core.find(find_all=True, idVendor=USBVendorIDs.USB_VID_ELGATO)
     for device in devices:
         try:
             # Check if it's a StreamDeck
             if device.idProduct in [
```
jeinstei, line 218: This should be a new PR and automated somewhat. Especially as it is StreamDeck specific. I'll make a new PR for that separately.

PR #551, opened 2026-02-09
Add based API for (fix #548) (for Autopage and other future applicaations)

### LINKED ISSUE #548: PR proposed, feedback/interest from devs? "Autopages"
(copied from the discord so I can get your advice/feedback) Hi ya'll,

So I wrote a little plugin that seems to basically work (the VSCode plugin).  But now I'm thinking of tackling something bigger (in a couple of months after I finish my current open-source task):

An easy way (probably toml or json) to textually describe (on a per app basis) easy automatic keybindings/buttons that will be added to the streamdeck when that app is in the foreground.  Mini featurelist/spec:

# Autopages
A way to **easily** automatically have pages/button behaviors without having to manually populate each page/button via the GUI. 

* Get informed by streamcontroller when the foreground app changes. 
* Based on window class/name regexes try to find a foo.ap.toml file for that app.  If found, parse that file and add buttons/actions.
* Autopage defs can be bunded into the app or in a toml-repo.  I'll create semi public 'master' autopages repo which anyone can send in PRs to to add new apps.  If you'd like we can someday move it into your StreamController gh org?
* I'll premake and include a set of common autopage def files (kwrite, vscode, etc..)
* Whenever the foreground app changes, if the user hasn't already specified a page for that app and and a ab.toml match is found -> create a page for that user.  Icons and labels will be auto selected for buttons based on regex matching vs the users installed asset packs.

This TBD autobuttons file will say stuff like:
*  "I'm for kwrite or vscode or gimp"
* Here's a list of buttons node (this node can have stuff like background color specified if the user wishes)
* Add a button with icon "copy" whos action is "paste ctrl-c"  (it will search <icon pack> for icons matching that regex). 
* Alternate more flexible ways of specifying custom icons could be provided in the future)
(The format will be very human readable and succinct - but it will be processed to generate json compatible with your existing import/export page format)

Useful?  Could this be done with current plugin api (I think probably no? if not I'm open to sending in a PR for the app)

geeksville:
# Summary of changes/motivation

The main component of this change is adding api.py, a dasbus based dbus API server with the following features:

* icon pack enumeration
* page enumeration
* addpage(name, json)
* removepage
* streamdeck enumeration
* export the datadir location (for eventual use adding other metadata/backgrounds on-the-fly)

To support this feature a small number of changes were needed elsewhere:

* Change add_page so it can throw if the page already exists (the threeish usages elsewhere in the app now have a try/catch to handle that)
* The flatpak yaml had to be updated to add dasbus (and since the tool is now newer it emitted slightly different yaml)

# How to use

This new api can be used either via the [autopage](https://github.com/geeksville/autopage) application or the "stream-client" test harness/client.  Run "pipx install stream-autopage" to get the current version of these tools.

Here's a [video](https://www.youtube.com/watch?v=nWYgs0SE4BA) of an app using the api (with streamcontroller automatically changing pages when apps change)

# Questions for devs

A few questions before I think this change should be merged:

* I think the autopage idea is pretty slick and shareable way to automatically add pages for detected running apps.  Please check out my [README](https://github.com/geeksville/autopage) for more info.  That said I tried to structure this (small) api so that it could eventually support other apps.
* I used dasbus because it is **much** newer than dbus-python (and no painful hand xml generation - it is mostly python attribute driven), the API is essentially a superset of dbus-python.  Elsewhere in your project you are using dbus-python.  Are you okay with both being in the app.  I think it is basically 'fine' but if you prefer I can amend this PR to use dbus-python (not my preference) or change your existing usages to use dasbus and remove dbus-python (my preference).  Your thoughts?
* This API is dbus based and I saw notes in the code with (early?) macos support.  dbus doesn't work there.  If macos is a requirement I could change this api to use some other message passing api (rabbitmq? zeromq? instead).  I'm not a huge fan of exposing this sort of 'user local stuff' via http though.  Thoughts?
* While making this change I noted a lot of 'dev time' dependencies are being included in the flatpak, I bet it could be made a lot smaller by doing a two stage build and not including those.  If you want me to look into this let me know.
* I've tried to structure the autopage app so that when it is a bit more mature we could just move it into the Streamcontroller flatpak (if ya'll want that).

Rock on, thanks for making this.  I'm happy to discuss this idea more either here or in #548


geeksville:
btw: i thought later on this - I bet you needed to leave in those 'devtime' things in the flatpak because you need to be able to load arbitrary python requirements.txt files for plugins.  So yeah - I can see how the container needs to be largeish.

geeksville:
also btw: I made the SetActivePage() call a method, but now that I know more about what dbus can do I think it really could have just been part of the property setter instead.  If you want I can remote that method and merge that code into the setter (which I think would simpilfy the api a bit)

Core447:
Thank you! Looks nice

geeksville:
oh awesome!  I'll release my app soon that uses this.  I need to update it.

Issue #334, opened 2025-02-11
BUG Crash on Import


My1:
**Describe the bug**
Basically when importing it seem to freeze, logs point to a font problem

**To Reproduce**
1. Delete ~/.var/app/com.core447.StreamController/data
2. Open Streamcontroller
3. Import

**Expected behavior**
it imports

**Screenshots**

![Image](https://github.com/user-attachments/assets/6af17e61-e16f-4798-8699-e782f4ac8ff1)

![Image](https://github.com/user-attachments/assets/943a6806-7722-4225-96cc-2c132b3b40c0)

**Additional context**
<details>
<summary> Logs, Serialnumer replaced with "NOPE"</summary>

```
File "/app/bin/StreamController/src/windows/PageManager/Importer/Importer.py", line 76, in import_from_streamdeck_ui
    ui_importer.perform_import()
    │           └ <function StreamDeckUIImporter.perform_import at 0x7fa6d5dc3920>
    └ <src.windows.PageManager.Importer.StreamDeckUI.StreamDeckUI.StreamDeckUIImporter object at 0x7fa6c479dac0>

  File "/app/bin/StreamController/src/windows/PageManager/Importer/StreamDeckUI/StreamDeckUI.py", line 102, in perform_import
    "font_family": font_family_from_path(self.export["state"][deck]["buttons"][page_name][button]["states"][export_state].get("font"))
                   │                     │    │               │                │          │                 └ '0'
                   │                     │    │               │                │          └ '4'
                   │                     │    │               │                └ '0'
                   │                     │    │               └ 'NOPE'
                   │                     │    └ {'state': {'NOPE': {'buttons': {'0': {'0': {'state': 0, 'states': {'0': {'text': '', 'icon': '', 'keys': '', 'write':...
                   │                     └ <src.windows.PageManager.Importer.StreamDeckUI.StreamDeckUI.StreamDeckUIImporter object at 0x7fa6c479dac0>
                   └ <function font_family_from_path at 0x7fa6d6462160>

  File "/app/bin/StreamController/src/windows/PageManager/Importer/StreamDeckUI/helper.py", line 12, in font_family_from_path
    font = ttLib.TTFont(path)
           │     │      └ '/usr/local/share/fonts/f/Font_Awesome_6_Pro_Regular_400.otf'
           │     └ <class 'fontTools.ttLib.ttFont.TTFont'>
[...]

PR #568, opened 2026-03-03
fix: StreamDeck UI importer - support v1 flat format 

henricook:
## Background

I was migrating from the original [streamdeck-ui](https://github.com/timothycrosley/streamdeck-ui) to StreamController. I know, I didn't even realise I was using the old old one. When importing my `~/.streamdeck_ui.json` backup, the importer window froze at "Importing..." with no error and no way to dismiss it.

The root cause is a format mismatch. The importer expects the `"states"` dict structure introduced in [streamdeck-linux-gui](https://github.com/streamdeck-linux-gui/streamdeck-linux-gui) v4.0.0 (config v2), where button properties are wrapped like:

```json
{"states": {"0": {"text": "Hello", "icon": "..."}}}
```

The original streamdeck-ui (config v1) stores properties flat on the button:

```json
{"text": "Hello", "icon": "..."}
```

With the v1 format, `button.get("states", {})` returns an empty dict so the state iteration loop never executes and every key is written as `{}` - the import "succeeds" but produces empty pages.

I also hit a second issue: my backup file was named `.streamdeck_ui.json.backup.20260301`. The `.json` extension check rejected it silently because `os.path.splitext` sees `.20260301` as the extension, and `show_error()` was a no-op.

While investigating I found two more latent bugs in the same code paths.

## What have I changed?

**1. Support streamdeck-ui v1 flat format** (`StreamDeckUI.py`)

When a button has no `"states"` key (or it's empty), the button data itself is now treated as state `"0"`. This is backward compatible - buttons that have the `"states"` dict still go through the existing code path.

**2. Implement `show_error()` and replace extension check with content validation** (`Importer.py`)

- `show_error()` was an empty no-op method. It now shows the error message in the progress bar and auto-closes after 3 seconds. I'm very happy to retract this if it's controversial(!)
- I've replaced the `.json` extension check with actual JSON parsing, so files with non-standard extensions (like `.json.backup.20260301`) work fine as long as they contain valid JSON.

**3. Fix `copy_asset` filename collision loop** (`AssetManagerBackend.py`)

The `while` loop that renames assets on filename collision never incremented its counter, so it would generate the same candidate filename repeatedly. I've also fixed the filename template which was re-splitting the already-modified name on each iteration, causing cumulative suffixes. I'll happily remove these if it complicates review.

**4. Add retry limit to `save_json`** (`StreamDeckUI.py`, `StreamController.py`)

Both importers have a `save_json` method that recursively retries on verification failure with no limit. I've added a cap of 3 retries.

## Testing

I've manually tested with a real streamdeck-ui v1 config file (6 configured buttons with icons, commands, and labels) and verified that:
- The import completes successfully with pages created containing the correct key data, icons, labels, and actions
- An invalid file shows an error message and the window auto-closes
- The existing `"states"` format still works as before

## Related

- #334 - similar import freeze (different root cause: missing font file, but a couple of puzzled people on there)

Here's my (lightly redacted) config:

[streamdeck_ui.json.backup.sanitized.zip](https://github.com/user-attachments/files/25721488/streamdeck_ui.json.backup.sanitized.zip)

### DIFF
--- a/src/backend/AssetManagerBackend.py
+++ b/src/backend/AssetManagerBackend.py
@@ -142,10 +142,11 @@ class AssetManagerBackend(list):
             dst_path = os.path.join(gl.DATA_PATH, "Assets", "AssetManager", "Assets", file_name)
         else:
             log.warning(f"File with same name already exists but sha256 does not match, renaming: {asset_path}")
+            original_base, ext = os.path.splitext(os.path.basename(asset_path))
             index = 2
             while file_in_dir(file_name, os.path.join(gl.DATA_PATH, "Assets", "AssetManager", "Assets")):
-                base, ext = os.path.splitext(file_name)
-                file_name = f"{base}-{str(index).zfill(2)}.{ext.replace('.', '')}"
+                file_name = f"{original_base}-{str(index).zfill(2)}{ext}"
+                index += 1
             dst_path = os.path.join(gl.DATA_PATH, "Assets", "AssetManager", "Assets", file_name)
 
         if asset_path == dst_path:
--- a/src/windows/PageManager/Importer/Importer.py
+++ b/src/windows/PageManager/Importer/Importer.py
@@ -13,6 +13,7 @@ You should have received a copy of the GNU General Public License
 along with this program. If not, see <https://www.gnu.org/licenses/>.
 """
 # Import gi
+import json
 import os
 import threading
 import gi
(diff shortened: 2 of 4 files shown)


Core447:
Looks very nice. Will go through it in detail in the next days and merge before the next release. Thanks for the great work!

Core447:
Amazing work!

henricook:
Ah amazing, cheers @Core447 !

PR #569, opened 2026-03-06
Improve Stream Deck+ touchscreen scheduling and preview coalescing

designgears:
## Summary

This changeset makes StreamController use the Stream Deck+ touchscreen API more effectively and removes a large amount of redundant UI-preview work.

On the hardware side, touchscreen updates now use partial-region scheduling instead of always treating the strip like a full-frame surface. On the UI side, preview updates are coalesced so rapid meter or volume changes do not flood GTK with stale thumbnail work.

## What changed

- Added `BetterDeck` context manager support in `BetterDeck.py` so StreamController can batch writes under the hardware library’s update lock.
- Reworked the media-player flush path in `DeckController.py`:
  - batch key and touchscreen writes under one deck lock
  - queue touchscreen work as region images instead of pre-encoded full-strip payloads
  - support partial touchscreen region uploads for dial updates
  - merge adjacent touchscreen regions before flushing
  - dedupe key and touchscreen writes by content hash
  - drop stale page writes instead of flushing them later
  - temporarily boost priority for real user interactions so volume/mute feedback wins over meter churn
- Added cached touchscreen base/current composition in `DeckController.py` so dial updates can replace only their own strip region.
- Fixed stale transparent pixels when no background is present by replacing the full dial slot during partial updates instead of alpha-pasting it back into the cached strip.
- Coalesced touchscreen preview updates before sending them into GTK in `DeckController.py`.

## Why

Before this change:
- live meters could crowd out more important strip updates
- dial updates rebuilt and resent more strip data than necessary
- transparent updates could leave stale pixels behind when no background was present
- the GTK preview path generated frequent stale-task abort logs during normal operation

After this change:
- interaction-driven updates are prioritized
- the Stream Deck+ strip uses partial-region writes
- redundant writes are skipped
- preview churn is reduced and the stale-task logs are gone


designgears:
Testing with my DeckWeaver plugin with meters enabled is an easy way to see the difference. Meters on and sound playing, the volume changes feel sluggish or unresponsive, with this change they feel native.

PR #460, opened 2025-07-24
feat: CLI commands for device management & state control

WakanaYuki:
This feature has been requested in this issue: #164 

### What’s new
- Device discovery  
  - `--list-devices`: shows ID, product name, serial, key layout, connection status; prints permission tips if access fails.  
  - `--list-pages`: lists every page, its items, and state counts.

- State control  
  - `--change-state SERIAL PAGE COORD STATE`: changes a single key’s state. Works via D-Bus if StreamController is running, or directly if not.

### Help Documentation
I added the new commands to the documentation, here is what the help functions now additionally states.
```
  --list-devices        List all connected StreamDeck devices and their properties
  --list-pages          List all available pages
  --change-state SERIAL PAGE COORDS STATE
                        Change the state of a StreamDeck item. Format: SERIAL PAGE COORDS STATE SERIAL: Device serial number (e.g., CL123456789) PAGE: Page name (e.g., Main, Soundboard) COORDS:
                        Position as x,y (e.g., 0,0 for top-left) STATE: State number to change to (0, 1, 2, etc.) Example: --change-state CL123456789 Main 0,0 1
```

### Files touched
- `globals.py` — CLI arg definitions, API request storage  
- `main.py` — listing commands, validation, API calls  
- `app.py` — D-Bus handler for `change_state`  
- `DeckController.py` — processes state-change requests at init

### Typical uses
- Check connected StreamDecks quickly  
- Script key-state changes 
  - thereby enabling companion applications to interact with StreamController item states
- Debug layouts and permissions

### Compatibility
Additive only; existing commands and APIs are untouched.

PR #570, opened 2026-03-07
Fix: use correct method name in on_change_state D-Bus handler

### LINKED ISSUE #571: Bug: --change-state CLI silently fails (calls non-existent method)
## Description

The `--change-state` CLI command (added in #460) silently fails — it exits with code 0 but has no effect on the StreamDeck device.

## Root Cause

`on_change_state` in `src/app.py` and the `api_state_requests` handler in `src/backend/DeckManagement/DeckController.py` both call `gl.page_manager.get_best_page_path_match_from_name()`, which does not exist on `PageManagerBackend`. The correct method is `find_matching_page_path()`, which `on_change_page` and the `api_page_requests` handler already use correctly.

The `AttributeError` is swallowed by GTK's D-Bus action handler, so there is no error output or log entry — the command just silently does nothing.

## Steps to Reproduce

1. Set up a StreamDeck button with multiple states
2. Run: `streamcontroller --change-state <SERIAL> <PAGE> <x,y> 1`
3. Observe: command exits successfully, but button state does not change
4. Check logs: no `on_change_state` log entry appears

## Expected Behavior

The button state should change on the physical device and in the UI, with a log entry like:
```
Successfully changed state of (x,y) to state 1 on device <SERIAL>
```

## Environment

- StreamController installed via AUR (`streamcontroller-git`)
- Stream Deck Original
- Arch Linux

## Fix

#570

tyvsmith:
## Summary

The `--change-state` CLI feature (added in #460) calls `gl.page_manager.get_best_page_path_match_from_name()`, but this method does not exist on `PageManagerBackend`. The correct method is `find_matching_page_path()`, which is what `on_change_page` and the `api_page_requests` handler already use correctly.

Two call sites are affected:
- `src/app.py` — `on_change_state` D-Bus handler (used when SC is already running)
- `src/backend/DeckManagement/DeckController.py` — `api_state_requests` handler (used when SC is not running and processes queued requests at startup)

The `AttributeError` is silently swallowed by GTK's action handler, so `--change-state` exits successfully but has no effect.

Fixes #571

## Fix

Replace `get_best_page_path_match_from_name` with `find_matching_page_path` in both call sites.

## Testing

Verified on a live StreamController instance (installed via AUR `streamcontroller-git`):

- **Before fix:** `streamcontroller --change-state AL08L2C53065 Desktop 4,1 1` exits silently with no effect, no log output
- **After fix:** state changes correctly on the physical Stream Deck, log confirms: `Successfully changed state of (4,1) to state 1 on device AL08L2C53065`

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -342,7 +342,7 @@ class App(Adw.Application):
             return
 
         # Find the requested page
-        page_path = gl.page_manager.get_best_page_path_match_from_name(page_name)
+        page_path = gl.page_manager.find_matching_page_path(page_name)
         if page_path is None:
             # Page not found - provide helpful suggestions
             available_pages = [os.path.splitext(os.path.basename(p))[0] for p in gl.page_manager.get_pages()]
--- a/src/backend/DeckManagement/DeckController.py
+++ b/src/backend/DeckManagement/DeckController.py
@@ -522,7 +522,7 @@ class DeckController:
             state = state_request["state"]
             
             # Get the page path for the specified page
-            requested_page_path = gl.page_manager.get_best_page_path_match_from_name(page_name)
+            requested_page_path = gl.page_manager.find_matching_page_path(page_name)
             
             if requested_page_path is None:
                 # Page not found - log available pages


Core447:
Thank you!

Issue #555, opened 2026-02-16
BUG: CPU usage increases when a label contains too many instances of the letter m


zedwick:
**Describe the bug**
CPU usage increases when using multiple lines in a single label and certain characters or number of characters. For example three lines with 8 characters will use ~15% CPU without scrolling, whilst three lines with 2 characters would use ~5% CPU. In some cases changing an `f` to an `m` triples the CPU usage, if the total character width happens to be at the threshold. Even new line characters count toward the total width of the label.

**To Reproduce**
1. Create a new blank page
2. On that new page choose a button and edit the center label
3. Fill the label with `m`s until just before it starts scrolling
4. Open up your favourite CPU monitor, I use `htop` in terminal. Find the StreamController process to monitor its CPU usage. Note its current usage (~5%)
5. With the line filled with `m`, add a new line character (can copy it by selecting the end of a line and the start of the next line in any text)
6. Observe the change in CPU usage (~11%)

**Expected behavior**
I would not expect such an increase in CPU usage for using too many characters without inducing scrolling.

**Additional context**
It seems to me that it is likely checking if the line would induce scrolling, and then doing something intensive in response. It must be counting the total character widths in the label, including the width of the new line character, and not taking into account any line breaks which actually make the block of text narrower.

The following does not increase CPU usage:

```
germ
fobs
frgg
```

Where-as the following will triple the CPU usage; note the change of `f` to `m` on the second line:

```
germ
mobs
frgg
```

The same is true for the following two examples, where the total width of the label is a little more clear:

```
germfobsfrgg


```

```
germmobsfrgg


```

Core447:
Should be fixed with #579. If not feel free to reopen this issue

Issue #123, opened 2024-05-31
[QUESTION] Should StreamController use this much RAM?


LordManhattan:
Fairly high RAM usage by StreamController. Is this caused by a memory leak, or is it by design? I have plenty of RAM so it's not affecting me personally, but I'm just wondering if it's by design or not. 

![Screenshot from 2024-05-31 13-47-23](https://github.com/StreamController/StreamController/assets/7291676/b82edfa8-cebb-4976-98f2-a9c85c2225af)

Core447:
Yeah, high memory usage is still a big problem. I think it is related to the actions as the consumption grows when adding actions and switching pages. I hope I can fix this soon

pnmtech:
I just realized to a much lesser extent that I have the same issue and with some short research I found m this https://github.com/flatpak/xdg-dbus-proxy/issues/51 and apparently the fix was found in https://github.com/sailfishos/xdg-dbus-proxy/blob/master/rpm/0001-Fix-GVariant-reference-leaks.patch I am not good with programming so I have no idea if my input was needed in the first place. I'm just trying to help. Really love the software. I've been enjoying using it

Core447:
@pnmtech Thank you for the links! However, I think the issue is caused by the app itself, because the same problem appears when using it outside of Flatpak

NightSprinter:
It isn't just the RAM usage, CPU as well, both Flatpak or otherwise.  Both get worse if you just make a button show the percentage of CPU or RAM used, and just press that button rapidly.  I also wonder if it's something in the underlying Python layer that StreamController uses?

Core447:
> CPU or RAM

The graphs or the "normal" ones? Because the graphs take up a huge amount of memory (that's why I added a warning in their config area).

I hope I can fix those issues soon, but it'll probably take some time to find the issues

Linuturk:
Wanted to chime in that I'm seeing this behavior as well, leading to resource contention on my system.

When you first start StreamController, memory usage is minimum.

![Image](https://github.com/user-attachments/assets/a130d49e-f95a-44ba-bf39-8cdc9516c118)

![Image](https://github.com/user-attachments/assets/7ab9f9ab-4b4b-4334-a7be-444d55369cad)

But over a day or two, you can see memory usage has exploded, and just gets worse over time.

![Image](https://github.com/user-attachments/assets/d797570a-96f4-4886-8fdd-a9d5b05f3aab)

Phrozenn1:
[...]

PR #579, opened 2026-03-17
fix: reduce idle CPU usage from ~25% to ~3%

### LINKED ISSUE #433: BUG: Notable increase in CPU usage
**Describe the bug**

In updating from ? to 1.5.0-beta.11 I've noticed streamcontroller now occupies more CPU than before when idle.

Previously I did not see StreamController at the top of the CPU usage on my desktop. Now tho:

```
    PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND                                                                                                                                                                                                               
2497339 coconnor  20   0 5901964 380748  94196 S   7.9   0.4   0:25.59 StreamControlle                                                                                                                                                                                                       
```

it's usually in first place.

My page only uses the audio controller widgets/actions. No animations. Static as far as I know.

**To Reproduce**

Upgrade to 1.5 from (?) I'll have to look it up

**Expected behavior**

Not having "too much" CPU usage.

**Screenshots**

**Additional context**

* 1.5.0-beta.11 from flatpak
* audio control 2.5.0

Not sure if this is specific to a plugin. Any pointers on how to profile a python flatpak application? Seems like something that should quickly give an answer if I knew how haha

oneandonlyno1:
## Summary

StreamController idles at ~25-30% CPU on KDE Wayland with flatpak, even with no user interaction. Profiling revealed three independent causes, all in `DeckController.py`:

- **KDE WindowGrabber** spawns 3 subprocesses (via `flatpak-spawn` + `kdotool`) every 200ms even when the active window hasn't changed. Now only fetches window name/class when the window ID actually changes (1 subprocess/cycle instead of 3).

- **MediaPlayer loop** runs at 30 FPS unconditionally, iterating all keys every tick. Now drops to 2 FPS when no animated content (video/GIF) is active and ramps back instantly when needed. Keys without video/GIF content are skipped entirely.

- **Plugin tick rendering** — plugins like CPUTemp call `set_center_label()` every tick with the same value, triggering full Pillow image renders (freetype font measurement + compositing + native format conversion) unnecessarily. Added label equality check to skip redundant renders, and an image hash guard in `ControllerKey.update()` to catch remaining cases.

Tested on CachyOS (KDE Plasma 6 Wayland) with Stream Deck MK.2 + XL (47 keys total), 12 plugins installed.

  Fixes #433
  Relates to #555
  Relates to #123

## Test plan

- [ ] Verify idle CPU drops significantly (~25% → ~3% on dual-deck setup)
- [ ] Verify GIF/video key animations still play at full FPS
- [ ] Verify background video wallpapers still animate
- [ ] Verify label scrolling still works for long text
- [ ] Verify plugin `on_tick` updates still appear (CPUTemp, MicMute, Weather)
- [ ] Verify key press responsiveness is unchanged
- [ ] Verify automatic page switching (KDE window grabber) still works

### DIFF
--- a/src/backend/DeckManagement/DeckController.py
+++ b/src/backend/DeckManagement/DeckController.py
@@ -1189,6 +1189,7 @@ class LabelManager:
         self.page_labels = {}
         self.action_labels = {}
         self.scroll_wait = 25
+        self._has_scroll_labels_cache: bool = None
 
         self.init_labels()
         self.frames: dict[str, dict[str, int]] = {
@@ -1213,6 +1214,7 @@ class LabelManager:
  
     def clear_labels(self):
         self.init_labels()
+        self._has_scroll_labels_cache = None
 
     def set_page_label(self, position: str, label: "KeyLabel", update: bool = True):
         if label is None:
@@ -1220,7 +1222,8 @@ class LabelManager:
             label.clear_values()
         else:
             self.page_labels[position] = label
-        
+
+        self._has_scroll_labels_cache = None
         if update:
             self.update_label(position)
 
@@ -1231,6 +1234,7 @@ class LabelManager:
         else:
             self.action_labels[position] = label
 
+        self._has_scroll_labels_cache = None
         GLib.idle_add(self.update_label_editor)
         if update:
             self.update_label(position)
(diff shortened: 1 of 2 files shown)


Core447:
Great work!

PR #595, opened 2026-05-11
fix: Add proper GIF support with transparency and per-frame timing

Dav1d-Fn:
## Problem

GIFs were decoded using OpenCV (`cv2.VideoCapture`), which has two limitations:

1. **No transparency support** — GIF frames in palette mode (`P`) with a transparency index were composited onto a white background, losing the alpha channel.
2. **Incorrect frame timing** — All frames played at a fixed FPS value, ignoring the per-frame delay stored in the GIF metadata. This caused animations with variable timing (e.g. a 2-second intro frame followed by fast animation frames) to play incorrectly or appear cut short.

## Solution

GIF files are now detected by extension and handled separately using PIL:

- **`key_video_cache.py`**: GIFs are opened with `Image.open()` and each frame is converted to RGBA, preserving transparency. Frames are cached as PNG instead of JPEG to retain the alpha channel. Per-frame delays are read from the GIF metadata (`img.info["duration"]`) and stored in `frame_delays`.
- **`KeyVideo.py`**: A new `_get_next_gif_frame()` method advances playback by accumulating elapsed milliseconds and comparing against the current frame's delay, rather than using a global FPS value.

All non-GIF video formats (mp4, webm, etc.) are unaffected and continue to use the existing cv2 code path.

## Testing

Tested with:
- Transparent GIF (palette mode with transparency index) — now renders correctly on dark backgrounds
- GIF with uniform frame delays — plays at correct speed
- GIF with variable frame delays (long first frame + short animation frames) — each frame now shown for its correct duration
- Non-GIF videos (mp4) — behaviour unchanged

### DIFF
--- a/src/backend/DeckManagement/DeckController.py
+++ b/src/backend/DeckManagement/DeckController.py
@@ -1156,12 +1156,11 @@ class BackgroundVideo(BackgroundVideoCache):
         region = (start_x, start_y, start_x + key_width, start_y + key_height)
         segment = image.crop(region)
 
-        # Create a new key-sized image, and paste in the cropped section of the
-        # larger image.
-        key_image = PILHelper.create_key_image(deck)
-        key_image.paste(segment)
-
-        return key_image
+        # Return the cropped segment directly, preserving alpha (matches
+        # BackgroundImage.crop_key_image_from_deck_sized_image above) instead
+        # of pasting onto an opaque RGB key image, which silently dropped
+        # transparency for GIF backgrounds.
+        return segment.convert("RGBA")
 
 class KeyGIF(SingleKeyAsset):
     def __init__(self, controller_key: "ControllerKey", gif_path: str, fps: int = 30, loop: bool = True):
--- a/src/backend/DeckManagement/Subclasses/KeyVideo.py
+++ b/src/backend/DeckManagement/Subclasses/KeyVideo.py
@@ -33,7 +33,14 @@ class InputVideo(SingleKeyAsset):
 
         self.active_frame: int = -1
 
+        # GIF timing state: track elapsed ms to honour per-frame delays
+        self._gif_elapsed_ms: float = 0.0
+        self._gif_last_tick: int = -1
+
     def get_next_frame(self) -> Image:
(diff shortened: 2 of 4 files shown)


Dav1d-Fn:
## Update: Performance optimization for GIF loading

After the initial fix landed I noticed that pages with multiple animated GIFs
still froze for several seconds on load (up to ~25 s with many large GIFs),
and switching back to a previously visited page repeated the full load.

This commit addresses that with three changes:

**Static preview first**
Frame 0 is now rendered synchronously in `__init__` (~5–15 ms per GIF).
Every button shows a static image immediately on page load; the background
thread then fills in the remaining frames and transitions to full animation
once ready.

**Per-frame delays cached to disk**
Delays are saved as `delays.json` alongside the PNG frames after the first
read. On subsequent starts the file is loaded instead of seeking through
every frame of the GIF — the slow path runs exactly once per GIF.

**Instance deduplication**
`VideoFrameCache.get_or_create(path, size)` returns the existing instance
for the same file+size combination. A GIF used on three buttons is opened
once instead of three times, and page switches that return to an already
loaded page are essentially free.

Dav1d-Fn:
## Update 2: Segfault fix and GIF timing improvements

**Segfault fix (key_video_cache.py)**

The previous update introduced a race condition: `__init__` ran outside the
global registry lock to avoid blocking during file I/O, but this allowed two
threads to create separate instances for the same GIF and write to the same
cache files simultaneously. PIL writes PNGs incrementally, so concurrent
writes to the same file corrupted the data and caused segfaults when the
file was later read back.

Fixed with a per-key lock: only one thread initialises a given (path, size)
at a time, while different keys remain fully parallel.

**GIF timing (KeyVideo.py)**

Two smaller timing bugs fixed:
- Replaced tick-based elapsed time with `time.perf_counter()` — playback
  no longer drifts when the media player is under CPU load
- Changed `if elapsed >= delay` to a `while` loop so the animation catches
  up correctly if multiple frames are due at once instead of falling further
  behind each tick

Core447:
Thank you! Great work!

PR #598, opened 2026-05-20
Fix semaphore leak

gensyn:
Running the git version of StreamController directly, it would very often crash for me with this or similar errors:
```
/usr/lib/python3.14/multiprocessing/resource_tracker.py:297: UserWarning: resource_tracker: There appear to be 8 leaked semaphore objects to clean up at shutdown: {'/mp-jvnwx97h', '/mp-cguo3zu8', '/mp-gkf96e_6', '/mp-go1e2_ae', '/mp-mnbqf7w1', '/mp-8yryfp80', '/mp-wsz2uv6z', '/mp-hhrjz222'}
  warnings.warn(
```

I told Copilot to investigate and it proposed this change; its reasoning:

>Fix: replace the multiprocessing.Process wrapper with a direct subprocess.Popen call in run_command() in src/backend/DeckManagement/HelperMethods.py.
>
>This eliminates the POSIX semaphore creation entirely for run_command calls. The detached subprocess is already fully independent of the parent process via start_new_session=True.

I did not have a single crash with this change and everything still works as I would expect. Additionally, I have a user of the HomeAssistantPlugin reporting crashes when changing pages which I suspect might be linked because the crash for me would often occur on page changes.

### DIFF
--- a/src/backend/DeckManagement/HelperMethods.py
+++ b/src/backend/DeckManagement/HelperMethods.py
@@ -16,7 +16,6 @@ from datetime import datetime
 from functools import lru_cache
 import hashlib
 from io import BytesIO
-import multiprocessing
 import os
 import subprocess
 import matplotlib.font_manager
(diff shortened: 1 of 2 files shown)


gensyn:
I now removed `shell=True` since it's a security risk. Copilot will keep finding things to comment on, but since this is only used to open a website in a browser, these are minor issues.
What I **really** would like to know is why this code fragment crashed my StreamController almost daily when I never even opened a website through it...

Core447:
Thanks! Can we keep `run_command` or will this cause issues?

gensyn:
If you want it, I can restore the old call order, but since `run_command` was only used by `open_web` or to otherwise open a website, I thought I might as well remove it.

gensyn:
So, what do you prefer?

Core447:
Right now all plugins are taking care of launching commands by themselves. If we keep `run_command` they could in theory rely on that instead of implementing it for each plugin

gensyn:
Ok, I re-introduced `run_command`.

gensyn:
Although I am not confident anymore that this entirely fixes the problem. I had these crashes again in the meantime - fewer though. I am still investigating.

Core447:
Merged it. If you find something new just open a new PR. Thanks for your great work!

PR #602, opened 2026-05-25
Add Touchypad ($15 StreamDeck like) (proof of concept, please discuss in issue)

geeksville:
Hi,

So I had a lot of fun in #551 though I understand if you aren't interested in merging it.  However I've been using it for the last few months and that got me into experimenting with "would it be possible to use these **very** cheap (but good) CYD devices" to make an open-source alternative to StreamDeck (with a built in high-end touchpad and advanced host-side scripting).

I looked at your implementation and it seemed to me the best way to offer my device/project to your users would be by subclassing from the abstract [StreamDeck](https://github.com/StreamController/streamcontroller-python-elgato-streamdeck/blob/master/src/StreamDeck/Devices/StreamDeck.py) class.  Which is what I did.  For the time being I'm using a nasty monkeypatch to run your StreamController binary with any attached 'TouchyDecks' added to the list of StreamDeck devices.

I use this nasty invocation from my developer shell: 
```bash
python -c "import touchy_bootstrap; touchy_bootstrap.main()"
``` 

It seems to work pretty well for an early proof of concept.  But before I get too far down this road I wanted to ask would you be interested in this PR if it was polished?  You can see a video of it working here: https://github.com/geeksville/touchy-pad#features-coming-soon 

If you are interested here's what I'm thinking as next steps.  cool?
* I'd slightly tweak the innards of https://github.com/StreamController/streamcontroller-python-elgato-streamdeck to add a public method to the API to add "register_controllers_factory(callback)", such a factory would be called at startup so that various other libraries could cleanly register new StreamDeck compliant shims.
* I'd use that API in the touchy-pad pypi project to register that it can make "StreamDeck" like shims.
* I'd make a small PR to your main app to add a dependency on touchy-deck and call that register function.

I hope you are having a nice Spring.
-Kevin

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -95,7 +95,7 @@ six==1.17.0
 smmap==5.0.2
 speedtest-cli==2.1.3
 streamcontroller-plugin-tools==2.0.2
-streamcontroller-streamdeck==0.1.5
+touchy-pad
 textual==2.1.2
 tinycss2==1.4.0
 tqdm==4.67.1
--- a/requirements.txt
+++ b/requirements.txt
@@ -95,7 +95,7 @@ six==1.17.0
 smmap==5.0.2
 speedtest-cli==2.1.3
 streamcontroller-plugin-tools==2.0.2
-touchy-pad
+streamcontroller-streamdeck==0.1.6
 textual==2.1.2
 tinycss2==1.4.0
 tqdm==4.67.1
--- a/touchy_bootstrap.py
+++ b/touchy_bootstrap.py
@@ -1,0 +1,98 @@
+"""Bootstrap shim: install the TouchyDeck enumerate hook, then run main.py.
+
+Imported (not executed directly) by the `streamcontroller-run` Justfile
+recipe via ``python -c "import touchy_bootstrap; touchy_bootstrap.main()"``.
+
+Honours these environment variables:
+
+* ``TOUCHY_SIM=1`` — also call ``create_sim_device(...)`` so a simulated
+  TouchyDeck appears in StreamController's device list.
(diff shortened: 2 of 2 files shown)


Core447:
Sure. Would love to support DIY devices - on the Discord there's even a channel for this because I planned to add support at some time. But would be great if you would do this.

Also merged #551

geeksville:
great!  I'm on vacation the next two weeks but when I get back I'll get right on it.  I'll also send in a finished touchpad-plugin that uses that API to appear as a 'streamdeck' in your app.  Since I opened this PR I used a similar approach to add a similar plugin to OpenDeck: https://www.reddit.com/r/touchypad/comments/1u0ykl3/touchypad_alpha_3_release_opendeck_is_solid_and

geeksville:
@Core447 okay - i'm starting work now by forking your https://github.com/StreamController/streamcontroller-python-elgato-streamdeck so I can add a register_controllers_factory(callback) method (per thread above).

But one question, I notice that that lib is a fork of some other lib that also has a fair number of recent changes.  Do you remember why you needed to fork?  If minor I might be able to instead just rebase to that so that hopefully these changes could go upstream someday.

Core447:
@geeksville We use our own version because we changed some stuff to get resume from sleep working properly.
You can see all changes here: https://github.com/abcminiuser/python-elgato-streamdeck/compare/master...StreamController:streamcontroller-python-elgato-streamdeck:master

Another reason was that abcminiuser was quite busy (and maybe still is - idk) and therefore it took long to get changes pushed to pypi through him.

Core447:
But the more is upstream the better

geeksville:
oh right!  good point about the delay with getting fixes to pypi being a
possible problem.  No worries!

PR #607, opened 2026-05-29
Fix: catch AssertionError in _overlay.resize()

akinam0:
Resolves an issue where the background feature would freeze after an AssertionError occurred where GIF or MP4 Files were used (Tested only mp4 and gif so far)

the addition of a try/except block with a transparent fallback image worked fine so far for me


akinam0:
New bug because of this fix occurred. No need to Pull for now

PR #617, opened 2026-06-12
Adding Restart option in tray.py and fixing Open button in Data Path

Mattdubs9699:
added an option to Quit and Restart
image


Core447:
Thank you!

Issue #566, opened 2026-03-03
fix: SIGSEGV ~7s after startup due to GTK threading violation in MediaPlayerThread


mattmattox:
## Symptom

The application crashes with a segfault approximately 7 seconds after startup:

```
2026-03-03 00:41:21.336 | SUCCESS  | src.app:on_activate:117 - Finished loading app
2026-03-03 00:41:28.738 | INFO     | streamcontroller_plugin_tools.BackendBase:on_disconnect:39 - Connection closed
/app/bin/launch.sh: line 3:     3 Segmentation fault (core dumped)
```

The `BackendBase:on_disconnect` log entries are a secondary effect — the rpyc backends detect the main process has died. They are not the root cause.

## Root Cause

`MediaPlayerThread` is a 30fps background thread that calls GTK widget methods directly without routing them through the GLib main loop. GTK is not thread-safe; calling widget operations from a non-main thread causes C-level memory corruption that manifests as SIGSEGV. Python exception handling cannot intercept these crashes.

The ~7 second delay matches the time it takes for the main window to be fully mapped and the media player to begin issuing image updates after startup.

## Affected Files

| File | Method | Line |
|---|---|---|
| `src/backend/DeckManagement/DeckController.py` | `ControllerKey.set_ui_key_image()` | ~2277 |
| `src/backend/DeckManagement/DeckController.py` | `ControllerTouchScreen.set_ui_image()` | ~2350 |
| `src/windows/mainWindow/DeckPlus/ScreenBar.py` | `ScreenBarImage.set_image()` | ~284 |

## Fix

Wrap each direct GTK widget call with `GLib.idle_add()` to schedule the update on the main loop thread:

```python
# DeckController.py — set_ui_key_image()
# Before:
self.deck_controller.get_own_key_grid().buttons[x][y].set_image(image)
# After:
GLib.idle_add(self.deck_controller.get_own_key_grid().buttons[x][y].set_image, image)

# DeckController.py — set_ui_image()
# Before:
screenbar.image.set_image(image)
# After:
GLib.idle_add(screenbar.image.set_image, image)

# ScreenBar.py — set_image()
# Before:
gl.app.main_win.sidebar.key_editor.icon_selector.set_image(dial_image)
# After:
GLib.idle_add(gl.app.main_win.sidebar.key_editor.icon_selector.set_image, dial_image)
```

[...]

PR #625, opened 2026-07-09
Fix GTK calls from non-main threads (same SIGSEGV class as #566)

CupOfOwls:
hey :)

I have been chasing a crash where StreamController dies with a silent SIGSEGV and no Python traceback, and found it comes from GTK4 widget calls made off the main loop thread. I believe this is the same failure class that #566 fixed for MediaPlayerThread; this PR covers three remaining sites:

1. `DeckManager.remove_controller()` is called from `FlatpakDeckDisconnectThread` / udev callbacks; `deck_stack.remove_page()` now goes through `GLib.idle_add`, mirroring how `add_newly_connected_deck()` already handles `add_page()`.
2. `MainWindow.show_info_toast()` has background-thread callers (e.g. `update_assets` in main.py); the `Adw.Toast` is now constructed and added on the main loop.
3. `MissingRow.hide_install_error()` runs on a `threading.Timer` thread; its four widget calls are now dispatched via `GLib.idle_add`.

I have been running these patches on my own machine (Stream Deck +, flatpak, on the beta.14 base) since early June and the disconnect-path crash has not recurred. Full disclosure, I did use Claude Code to help track this down and write the fix.

Happy to split this into separate PRs or adjust anything, just let me know :)

### DIFF
--- a/src/backend/DeckManagement/DeckManager.py
+++ b/src/backend/DeckManagement/DeckManager.py
@@ -205,7 +205,11 @@ class DeckManager:
     def remove_controller(self, deck_controller: DeckController) -> None:
         self.deck_controller.remove(deck_controller)
         if recursive_hasattr(gl, "app.main_win.leftArea.deck_stack"):
-            gl.app.main_win.leftArea.deck_stack.remove_page(deck_controller)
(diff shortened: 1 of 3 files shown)


Core447:
Thank you!

PR #626, opened 2026-07-09
Hold Gio.Application while keep-running is enabled (silent idle-quit in background mode)

CupOfOwls:
hey :)

With keep-running enabled, closing the main window only hides it, and Gio.Application's reference counting can then decide the app is idle and quit the process silently. The deck goes dark with no traceback and no log line, which makes it look like a crash.

This holds the application in `__init__` while keep-running is active and releases it in `on_quit()`, so the lifecycle is pinned to the explicit quit path. I believe #558 reworks this area more broadly; this is meant as the minimal bugfix for current main, and I am happy to close it if that lands first.

I have been running this since late May with the app in background mode daily and have not seen the silent quit since. Full disclosure, I did use Claude Code to help with this.

Let me know if you would like anything changed :)

### DIFF
--- a/src/app.py
+++ b/src/app.py
@@ -58,6 +58,20 @@ import globals as gl
 class App(Adw.Application):
     def __init__(self, deck_manager, **kwargs):
         super().__init__(**kwargs)
+        # Hold the application when running with keep-running enabled:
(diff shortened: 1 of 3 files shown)


Core447:
Thank you!

PR #632, opened 2026-07-26
add logind lock screen detection for Niri

nicosemp:
The current lock screen detection doesn't support [Niri](https://github.com/niri-wm/niri).

This PR adds support for it by trying lock screen detection via logind, This should also work on Gnome or KDE, but their specific detection is left untouched as a fallback.

Tested locally on NixOS with an overlay [overlay-test-diff.md](https://github.com/user-attachments/files/30380808/overlay-test-diff.md), basically replacing these 3 files before rebuilding.


Core447:
Great work! Thank you!

PR #633, opened 2026-07-27
Fix page import and cached-page reload callbacks

Oneiros667:
## Summary

- use the validated import/duplicate callback name when adding a page
- fix the `allow_reload` keyword used for active cached-page reloads

## Problem

`import_page_name_selected_callback` currently reads its local `page_path`
before assigning it. Both page import and duplicate reach that callback, so
they fail with `UnboundLocalError`.

`reload_pages_with_path` passes `allow_relaod` to
`DeckController.load_page`, whose supported parameter is `allow_reload`.
Reloading an active cached page therefore fails with `TypeError`.

## Verification

- both changed modules compile
- page import callback regression harness preserves the validated name, JSON
  document, selector-row update, and `PageAdd` signal
- cached-page regression harness preserves the active-page guard and all reload
  flags while passing `allow_reload=True`

### DIFF
--- a/src/backend/PageManagement/PageManagerBackend.py
+++ b/src/backend/PageManagement/PageManagerBackend.py
@@ -337,7 +337,7 @@ class PageManagerBackend:
             if page.deck_controller.active_page != page:
                 continue
 
-            page.deck_controller.load_page(page, allow_relaod=True,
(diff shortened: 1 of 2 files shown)


Core447:
Thank you!

PR #634, opened 2026-07-30
Don't attach dbus-python to the GLib main loop (fixes random SIGABRT in dbus_watch_handle)

jslay88:
## Problem

StreamController dies with `SIGABRT` (exit 134) after anywhere from hours to days of uptime. Nothing lands in `logs.log` because the abort happens in native code. The journal has it:

```
dbus[3]: arguments to dbus_watch_handle() were incorrect, assertion "watch != NULL" failed in file ../dbus/dbus-watch.c line 738.
This is normally a bug in some application using the D-Bus library.
/app/bin/launch.sh: line 3:     3 Aborted                 (core dumped) python3 /app/bin/StreamController/main.py "$@"
app-com.core447.StreamController@autostart.service: Main process exited, code=exited, status=134/n/a
app-flatpak-com.core447.StreamController-2233223863.scope: Consumed 1h 5min 6.185s CPU time over 5d 4h 38min 11.318s wall clock time
```

Stack from the core (1.5.0-beta.15 flatpak, KDE/Wayland):

```
#3  libdbus-1.so.3.38.3            <- assertion + abort
#4  libdbus-1.so.3.38.3
#5  libdbus-1.so.3.38.3            <- dbus_watch_handle
#6  _dbus_glib_bindings...so       <- io_handler_dispatch
#7  libglib-2.0.so.0               <- g_main_dispatch
#8  libglib-2.0.so.0
#9  libglib-2.0.so.0
#10 libgio-2.0.so.0
#11 libffi.so.8
```

## Cause

dbus-python's GLib integration is not thread safe, which is longstanding and [documented by the dbus maintainer](https://lists.freedesktop.org/archives/dbus/2014-December/016467.html) ("dbus-glib ... makes no attempt to be thread-safe"). In `dbus-gmain.c`:

```c
static void
watch_toggled (DBusWatch *watch, void *data)
{
  /* Because we just exit on OOM, enable/disable is
   * no different from add/remove
   */
  if (dbus_watch_get_enabled (watch))
    add_watch (watch, data);
  else
    remove_watch (watch, data);
}
```

libdbus toggles the write watch on essentially every message send, on whichever thread made the call, so both paths run off the main thread. They mutate an unlocked `GSList` on the connection (`cs->ios = g_slist_prepend (...)` in `connection_setup_add_watch`, `handler->cs->ios = g_slist_remove (...)` in `io_handler_destroy_source`). Meanwhile the main loop dispatches those same sources, and `io_handler_dispatch` has no NULL check:

```c
  /* Note that we don't touch the handler after this, because
   * dbus may have disabled the watch and thus killed the
   * handler.
   */
  dbus_watch_handle (handler->watch, dbus_condition);
```

`io_handler_watch_freed` is what sets `handler->watch = NULL`.

`main.py` called `DBusGMainLoop(set_as_default=True)`, which attaches every `dbus.SessionBus()` created afterwards, including the shared singleton plugins get. Plugins then call D-Bus from the per-deck `tick_actions` threads via `on_tick()`, roughly once a second, forever. `MediaPlugin` and `Battery` both do. That is a continuous race against the main loop, and eventually it loses.

Note this is not the older `dbus_threads_init_default()` problem. libdbus has been thread-safe by default since 1.8 and the runtime ships 1.16.2, so `threads_init()` is a no-op here. The unsynchronised state is dbus-python's own GLib glue.

## Fix

Stop attaching any dbus-python connection to the main loop. Without the attachment there are no GSources, `io_handler_dispatch` never runs, and the assertion is unreachable, which fixes it for plugin code too without plugins needing to change.

dbus-python is still used for blocking method calls (`quit_running`, `make_api_calls`, `GnomeExtensions`), which never needed a main loop. The four places that needed signals move to GDBus, which is thread safe and already used in this codebase (`src/backend/trayicon.py` uses `Gio.DBusProxy` / `Gio.bus_get_sync`, `src/api.py` uses dasbus).

* `main.py`: drop `DBusGMainLoop(set_as_default=True)` and the unused `import dbus.service`
* `LockScreenDetector`: add `subscribe_to_screen_saver()` on top of `Gio.DBusConnection.signal_subscribe`
* `Detectors/{KDE,Gnome,Cinnamon}`: use it instead of `add_signal_receiver`, and stop re-installing the main loop from the `LockScreenManager` setup thread
* `WindowGrabber/Integrations/Gnome`: port to `Gio.DBusProxy` and its `g-signal`
* `tests/`: stdlib regression guard, no new dependency

## Compatibility

`add_signal_receiver` is the only dbus-python API that breaks without a default main loop (blocking calls are unaffected, verified in the flatpak). I checked all 59 plugins in the official store by cloning each and scanning for `add_signal_receiver`, `connect_to_signal`, `DBusGMainLoop`, `dbus.mainloop`, `reply_handler`, `dbus.service.Object` and `set_default_main_loop`.

**Zero hits.** Only three touch dbus-python at all, and all three make blocking calls only, so they keep working unchanged:

| Plugin | Bus | Usage | Called from |
| --- | --- | --- | --- |
| MediaPlugin | session | MPRIS `get_object` + `Interface().Get()` | `on_tick` |
| Battery | system | UPower `get_object` + `Interface()` | `on_tick` |
| GnomeWindowCalls | session | Gnome Shell Extensions `get_object` | elsewhere |

None use pydbus, dasbus, sdbus or jeepney either. Two of those three are currently exposed to this crash and are fixed by this change.

The only remaining exposure would be an out-of-store plugin receiving dbus-python signals. It would now get a clear `RuntimeError` from dbus-python instead of racing, and it was already crash-prone. Happy to add an opt-in helper for that case if you would prefer one.

## Testing

Attached gdb to a process running the old configuration and broke on `dbus_watch_handle`. The live backtrace matches the crash core frame for frame:

```
#0  dbus_watch_handle ()                        libdbus-1.so.3
#1  ??                                          _dbus_glib_bindings...so
#2  g_main_dispatch ()                          libglib-2.0.so.0
#3  g_main_context_iterate_unlocked.isra ()     libglib-2.0.so.0
#4  g_main_loop_run ()                          libglib-2.0.so.0
#5  ffi_call_unix64 ()                          libffi.so.8
rdi            0x55c0448f77c0
```

`rdi` is the `DBusWatch *`. Valid there, `0` in the crash.

With the patch the library backing frame #1 is never loaded, so that frame cannot exist:

```
$ grep -c _dbus_glib_bindings /proc/<pid>/maps
5     # 1.5.0-beta.15 as shipped
0     # this branch
```

Ran the patched app on KDE/Wayland. Decks, pages, plugins and the dasbus API all load normally, and `dbus-monitor` shows the lock screen detector registering a match rule identical to the old `add_signal_receiver` one:

```
type='signal',interface='org.freedesktop.ScreenSaver',member='ActiveChanged',path='/org/freedesktop/ScreenSaver'
```

`python3 -m unittest discover -s tests -t .` passes, and fails on 129bdb5 flagging all four signal sites.

Two things I could not cover. I never got the abort to fire on demand: the stress tool ran 15 minutes and about 9M calls without hitting it, which is consistent with a race that took 5d 4h in the wild, so `tests/manual_dbus_thread_stress.py` is a manual tool rather than a test. And since I am on KDE, the Gnome window grabber port and the Gnome/Cinnamon detectors are reviewed but not exercised at runtime. Would appreciate a second pair of eyes from someone on GNOME or Cinnamon.


Core447:
Thank you! Great work!!

PR #640, opened 2026-08-22
fix: Treat a title-less window as a window change on KDE

mhlas7:
get_title() rejected any title under two characters, and the desktop window (plasmashell) has an empty title. Minimising every window therefore emitted no active-window-changed event at all, so pages configured to auto-change never fell back to the default page once a match had been made.

Accept an empty title, but re-query once first: a window that is still mapping also reports an empty title briefly, and treating that transient state as the desktop would load the wrong page.


Core447:
Thanks!

PR #641, opened 2026-08-22
Add window grabber support for mangowm

M-Pistillucci:
### Description
Adds native window tracking support for the MangoWM compositor, based on the already present implementation for hyprland.

### Testing
I've tested locally the event stream over Unix socket on MangoWM and everything seems to be working, as well as the polling fallback


Core447:
Great work! Thank you!

